PR #1, opened 2020-06-13
Asynchronous service design

rubdos:
Hi!

as I already said [in a comment](https://github.com/Michael-F-Bryan/libsignal-protocol-rs/issues/40#issuecomment-636543016), I'm rebuilding a Signal app for SailfishOS, called [Whisperfish](https://gitlab.com/rubdos/whisperfish/).

This version makes use of Qt and Actix, which [share their event loop](https://www.rubdos.be/corona/qt/rust/tokio/actix/2020/05/23/actix-qt.html).  If I'm going to contribute to this service crate (which I intend to!), the interfaces should be async-compatible.

This is a first experiment to define the AccountManager as a trait (since the runtime is an unknown, also in a synchronous context), and makes the `request_sms_verification_code` method asynchronous (it yields the `Future` when the request has been sent). To make `async` methods in a trait, I've opted to use the [`async-trait` crate](https://crates.io/crates/async-trait), because currently `async` `trait`s are not implemented (and are [hard to implement](https://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/), although not impossible).

Note that the intention here is to keep the base `-service` crate *runtime agnostic*. As a next step, I would provide a Tokio/Actix based implementation of the `trait`s defined in the `-service` crate.

What do you think?

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -1,63 +1,17 @@
-use crate::{TrustStore, USER_AGENT};
+use crate::push_service::{PushService, SmsVerificationCodeResponse};
+
 use failure::Error;
 
-pub struct AccountManager;
-
-impl AccountManager {
-    pub fn builder() -> AccountManagerBuilder {
-        AccountManagerBuilder::default()
-    }
-
-    pub fn request_sms_verification_code(&mut self) -> Result<(), Error> {
-        unimplemented!()
-    }
+pub struct AccountManager<Service> {
+    service: Service,
 }
 
-pub struct AccountManagerBuilder {
-    username: Option<String>,
-    password: Option<String>,
-    user_agent: String,
-    server: Option<String>,
-    trust_store: Option<TrustStore>,
-}
+impl<Service: PushService> AccountManager<Service> {
+    pub fn new(service: Service) -> Self { Self { service } }
 
-impl AccountManagerBuilder {
-    pub fn credentials<U: Into<String>, P: Into<String>>(
+    pub async fn request_sms_verification_code(
         &mut self,
-        username: U,
-        password: P,
-    ) -> &mut Self {
-        self.username = Some(username.into());
-        self.password = Some(password.into());
-        self
-    }
-
-    pub fn server<S: Into<String>>(&mut self, server: S) -> &mut Self {
-        self.server = Some(server.into());
-        self
-    }
-
-    pub fn trust_store(&mut self, trust_store: TrustStore) -> &mut Self {
-        self.trust_store = Some(trust_store);
-        self
-    }
-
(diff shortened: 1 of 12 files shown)


rubdos:
I'm gonna go a bit around, see what I need for my app and implement that, and then see how it can be cleanly abstracted from to allow different runtimes, possibly even synchronous ones.

libsignal-service/src/account_manager.rs
```diff
@@ -1,14 +1,32 @@
...
+    push_service::SmsVerificationCodeResponse, TrustStore, USER_AGENT,
+};
+
+use async_trait::async_trait;
 use failure::Error;
 
-pub struct AccountManager;
+#[async_trait(?Send)]
+pub trait AccountManager {
+    fn builder() -> AccountManagerBuilder { AccountManagerBuilder::default() }
```
Michael-F-Bryan, line 10: 90% of the time you don't want to specify how something is constructed as part of a trait. That's because now you're contractually obliged to only ever use the information inside `AccountManagerBuilder`... For example, what if I need to pass in a the mailbox to another actor, or a reference to an event loop (tokio or Qt) that I can spawn callbacks on?

Instead of the `builder()` and `from_builder()` methods I'd leave it up to the concrete type to determine how it is constructed, and use trait methods to only define required behaviour (e.g. `request_sms_verification_code()`).

Michael-F-Bryan:
> Note that the intention here is to keep the base -service crate runtime agnostic. As a next step, I would provide a Tokio/Actix based implementation of the traits defined in the -service crate.

I like this idea. The `libsignal-service` crate shouldn't care which runtime you are using, instead requiring a handful of low-level components that we can build (e.g. a way to send requests to the signal servers) and the logic for account management and everything else can be built on top of that.

rubdos:
I squashed a bit because of the experimental design. I'm merging this into master either today or tomorrow, feel free to comment away!

libsignal-service/src/account_manager.rs
rubdos: I've gotten rid of the forced constructors, but that also means the `AccountManagerBuilder` was kinda out of place (since the AccountManager is now built from a `PushService`).

Michael-F-Bryan:
I like where this is going :+1:

Is the long-term plan for `PushService::get()` to return some sort of [`Response`](https://docs.rs/http/0.2.1/http/response/struct.Response.html) with the contents of the GET request?

And likewise, `MessageReceiver` feels like some sort of `futures::Stream` backed by a websocket.

rubdos:
Absolutely and exactly the plan! MessegeSender would likewise become a sink. This way it mirrors Java's version about exactly, while staying true to the async spirit!

PR #5, opened 2020-07-11
Add the two group flags from textsecure

mjtorn:
And they are u32 there, but Whisperfish uses i32
so let's align with that.


rubdos:
lgtm!

PR #9, opened 2020-07-13
impl ServiceCipher, AttachmentCipher, keep alive and CDN fetching

rubdos:
- [x] Depends on resolution of https://github.com/Michael-F-Bryan/libsignal-protocol-rs/issues/59, e.g. https://github.com/Michael-F-Bryan/libsignal-protocol-rs/pull/60
- [ ] decryption
  - [x] `PrekeyBundle`, add new prekeys
  - [x] `Ciphertext`
  - [x] ~~I regularly get `InvalidFrameError { reason: "Envelope has unknown type." }`, find out what I'm missing.~~ Envelope with type Receipt has to be *handled by the caller*.
  - [ ] ~~`UnidentifiedSender`~~ postponed to #10 
- [ ] ~~Encryption~~ postponed until [Whisperfish alpha.3](https://gitlab.com/rubdos/whisperfish/-/milestones/6).
- [x] Fetch asynchronously from CDN
- [x] Keep alive timer


rubdos:
@mjtorn if you feel like it, please let me know what you think about 7da7702.

libsignal-service/src/models.rs
```diff
@@ -29,6 +28,6 @@ pub struct Message {
     pub message: String,
     pub attachments: Vec<Attachment<u8>>,
     pub group: Option<Group>,
-    pub timestamp: i64,
-    pub flags: i32,
+    pub timestamp: u64,
+    pub flags: u32,
```
mjtorn, line 32: As stated elsewhere, I can't remember the story behind how the types came to be; there was something from Go and something ended up in the Diesel models, but I don't think either one of us consulted the protobufs until now.

Should make sense to use those as a starting point and if everything compiles cleanly from here on, that's just winning.

mjtorn: Tested current Whisperfish against this and there is some work to do to get it to compile. But if this is what OWS is sending in, it's what should be, and the code adapt to it.

mjtorn: Looking at the errors and thinking back, this would have been a side-effect of [Diesel and the db not using unsigned ints](https://users.rust-lang.org/t/diesel-use-of-derive-insertable-causes-a-compile-error/13838/5).

It's a good guess the Go version of Whisperfish does not care about this, because the database has no unsigned types yet the Go code never complained.

Doubt there's more to say here; I'll adapt Whisperfish's use of Diesel around this.

libsignal-service/src/cipher.rs
```diff
@@ -0,0 +1,210 @@
...
+        return Ok(());
+    }
+
+    let mut start_of_padding = 0;
+    for (i, b) in contents.iter().enumerate().rev() {
+        if *b == 0x80 {
+            start_of_padding = i;
+            break;
+        }
+        if *b != 0x00 {
```
mjtorn, line 203: Just the smallest nit to pick: this could still be `else if` though the previous `if` breaks the loop on a condition where `*b != 0x00`.

rubdos: Maybe best for being consistent with the Java version. I haven't really looking into *what* the loop does exactly, so I didn't really Rustify it decently... just mimiced the Java thing.

mjtorn: I think it also helps readability, and though modern compilers probably know this stuff, I always like things explicit. Java has `else if` as well, though Java should also know how to compile that properly.

rubdos:
@Michael-F-Bryan this includes quite a bit of crypto, would you mind giving it a read and then a seal-of-approval?

libsignal-service/src/attachment_cipher.rs
```diff
@@ -0,0 +1,147 @@
...
+        Ok(())
+    }
+
+    #[test]
+    fn attachment_encrypt_decrypt_bad_key() -> Result<(), Error> {
+        let mut key = [0u8; 64];
+        let mut iv = [0u8; 16];
+        rand::thread_rng().fill_bytes(&mut key);
+        rand::thread_rng().fill_bytes(&mut iv);
+        let plaintext = b"Peter Parker";
```
shekohex, line 134: 😉

rubdos: https://github.com/signalapp/libsignal-service-java/blob/master/java/src/test/java/org/whispersystems/signalservice/api/crypto/AttachmentCipherTest.java#L27 :-)

libsignal-service/src/messagepipe.rs
```diff
@@ -57,72 +70,204 @@ impl<WS: WebSocketService> MessagePipe<WS> {
...
+            ..Default::default()
+        };
+        let mut buffer = BytesMut::with_capacity(msg.encoded_len());
+        msg.encode(&mut buffer).unwrap();
+        self.ws.send_message(buffer.into()).await?;
+        log::trace!("request on route.");
+        Ok(())
+    }
+
+    /// Send request and returns the response as a *nested* Future.
```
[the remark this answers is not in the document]
rubdos: Sending the request required borrowing `self`, while receiving did not. This decoupled the lifetimes. Maybe there is a cleaner solution though.

libsignal-service/src/cipher.rs
```diff
@@ -0,0 +1,212 @@
...
+        return Err(ServiceError::InvalidFrameError {
+            reason: format!("Unknown version {}", version),
+        });
+    } else if version == 2 {
+        // No-op
+        return Ok(());
+    }
+
+    let mut start_of_padding = 0;
+    for (i, b) in contents.iter().enumerate().rev() {
```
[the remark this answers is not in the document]
rubdos: TIL!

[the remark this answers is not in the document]
rubdos: Fixed in 19dff83e

libsignal-service/src/proto.rs
```diff
@@ -35,7 +35,7 @@ impl WebSocketRequestMessage {
             }
         }
 
-        false
+        true
```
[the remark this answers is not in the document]
rubdos: This was basically copied badly from the Java implementation.

libsignal-service/src/messagepipe.rs
```diff
@@ -57,72 +70,204 @@ impl<WS: WebSocketService> MessagePipe<WS> {
...
+            verb: Some("GET".into()),
+            ..Default::default()
+        };
+        let request = self.send_request(request).await?;
+        Ok(async move {
+            let response = request.await?;
+            if response.status() == 200 {
+                Ok(())
+            } else {
+                log::warn!(
```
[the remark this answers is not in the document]
rubdos: 907ce23a

libsignal-service/src/attachment_cipher.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    // Get rid of IV and MAC
+    ciphertext.copy_within(16..(plaintext_len + 16), 0);
+    ciphertext.truncate(plaintext_len);
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
rubdos: Done. Made me find a bug in Whisperfish; apparently you still have to truncate the plaintext to the length given in the `AttachmentPointer`.  Maybe that should be done in the `decrypt_in_place` function too...

PR #12, opened 2020-08-14
Secondary device registration

gferon:
This adds support for secondary device provisioning from an existing Signal device, similar to what's used when setting up Signal-Desktop.

TODO:
- [ ] Initial review and design/code-style discussion
- [x] Decide whether to reuse `ServiceError` or keep the separate error type
- [ ] Fix the issue when the public key contains characters that needs to be URL encoded for the QR code :exploding_head:
- [x] Remove custom git URL in `Cargo.toml` once the PRs in `libsignal-protocol-rs` are reviewed and merged

Depends on the following PRs which should be merged first:
- [x] Michael-F-Bryan/libsignal-protocol-rs#63
- [x] Michael-F-Bryan/libsignal-protocol-rs#62
- [ ] Michael-F-Bryan/libsignal-service-rs#9


libsignal-service/src/configuration.rs
```diff
@@ -33,3 +34,46 @@ impl Credentials {
...
+            ],
+            cdn_urls: vec![
+                "https://cdn-staging.signal.org".into(),
+                "https://cdn2-staging.signal.org".into(),
+            ],
+            contact_discovery_url: vec![], // TODO: add this one
+            certificate_authority: ROOT_CA.into(),
+        }
+    }
+}
```
rubdos, line 56: I'm  still not sure whether we want these in the library, what do you think @Michael-F-Bryan, @shekohex?

libsignal-service/src/provisioning.rs
```diff
@@ -0,0 +1,328 @@
...
+    }
+}
+
+#[pin_project]
+pub struct ProvisioningPipe<WS: WebSocketService> {
+    ws: WS,
+    #[pin]
+    stream: WS::Stream,
+    provisioning_cipher: ProvisioningCipher,
+}
```
rubdos, line 118: Maybe this should be a state machine with types for every step, to make it very Rusty.

rubdos: It can pass the stream, cipher and ws through every step, but the type encodes the current step in the process.

libsignal-service/src/push_service.rs
```diff
@@ -137,6 +172,25 @@ pub trait PushService {
...
+        Ok(VoiceVerificationCodeResponse::CallIssued)
+    }
+
+    async fn confirm_device(
+        &mut self,
+        confirm_code: u32,
+        confirm_code_message: &ConfirmCodeMessage,
+    ) -> Result<DeviceId, ServiceError> {
+        self.put(
+            dbg!(&format!("{}{}", DEVICE_PATH, confirm_code)),
```
rubdos, line 188: Stale `dbg!` (didn't  know about its existence, TIL), does `log` have an equivalent?

gferon: it's really not important, I'll just remove it.

libsignal-service-actix/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2018"
 
 [dependencies]
 libsignal-service = { path = "../libsignal-service" }
-libsignal-protocol = { git = "https://github.com/Michael-F-Bryan/libsignal-protocol-rs" }
+libsignal-protocol = { path = "../../libsignal-protocol-rs/libsignal-protocol" }
```
boxdot, line 11: This line and the corresponding line below need to be reverted to be able to compile.

Cargo.toml
```diff
@@ -3,3 +3,6 @@ members = [
     "libsignal-service",
     "libsignal-service-actix",
 ]
+
+[patch.crates-io]
```
boxdot, line 7: Instead of patching crates.io, I guess you wanted to patch "https://github.com/Michael-F-Bryan/libsignal-protocol-rs"

gferon: I just fixed it, thanks.

libsignal-service/src/provisioning.rs
Michael-F-Bryan: Does using a state machine give us anything in terms of correctness (by design the public API can't be used incorrectly) or make the code simpler? If not, I wouldn't bother because it'll add more work and increase cognitive load (instead of needing to understand a single type with a couple methods, you need to mentally parse a state machine and its various transitions).

libsignal-service/src/configuration.rs
Michael-F-Bryan: I imagine having access to Signal's certificates and relevant URLs will make getting started a lot simpler. Are there any security or copyright issues with embedding the certificate?

If you don't feel comfortable endorsing Whisper Systems as the de-facto standard, an alternative to `Default` would be putting it in a `ServiceConfiguration::with_signal_servers()` constructor.

rubdos: I think I would prefer that indeed.

libsignal-service/src/provisioning.rs
```diff
@@ -0,0 +1,328 @@
...
+        }
+
+        use aes::Aes256;
+        // libsignal-service-java uses Pkcs5,
+        // but that should not matter.
+        // https://crypto.stackexchange.com/questions/9043/what-is-the-difference-between-pkcs5-padding-and-pkcs7-padding
+        use block_modes::{block_padding::Pkcs7, BlockMode, Cbc};
+        let cipher = Cbc::<Aes256, Pkcs7>::new_var(&parts1, &iv)
+            .expect("initalization of CBC/AES/PKCS7");
+        let input = cipher.decrypt_vec(cipher_text).expect("decryption");
```
rubdos, line 106: Maybe being a bit pedantic here, but I think this should be "easy" to decrypt in-place, such that we don't need another allocation.  I probably did the same elsewhere though, where it matters more.

rubdos (commented):
Some minor remarks, otherwise looking good.

libsignal-service/protobuf/Provisioning.proto
```diff
@@ -10,6 +10,10 @@ package signalservice;
 option java_package         = "org.whispersystems.signalservice.internal.push";
 option java_outer_classname = "ProvisioningProtos";
 
+message ProvisioningUuid {
+  optional string uuid = 1;
+}
+
```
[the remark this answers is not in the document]
gferon: {{libsignal-service-java}} does not have this feature, hence it's not in the `.proto` files. I got this from https://github.com/Turasa/libsignal-service-java/blob/master/service/src/main/proto/Provisioning.proto

gferon:
> Some minor remarks, otherwise looking good.

thanks for the comments, I'll check them out and fix the code tomorrow!

rubdos:
Could you point this to master instead? I've merged `servicecipher`.

libsignal-service/src/provisioning.rs
gferon: Indeed, I reused the code from the message receiving function.

PR #13, opened 2020-09-14
Secondary device registration

gferon:
This adds support for secondary device provisioning from an existing Signal device, similar to what's used when setting up Signal-Desktop.

This is a new iteration of #12 targeting the `master` branch, which brings the registration logic inside of the `libsignal-service-actix` crate and not in the example. The example now only takes care of generating the QRcode and saving the result to a bunch of files.

TODO:
- [x] Once/if design is accepted by @rubdos, introduce a new error type in `libsignal_service_actix::provisioning` to not use `failure::Error`


rubdos:
Postponing the error type!

PR #14, opened 2020-09-18
rustfmt with stable

gferon:
The difference is very small and people generally develop using stable these days.


rubdos:
@Michael-F-Bryan if you're okay with using rustfmt-stable, let's merge this.

PR #19, opened 2020-10-21
Use new libsignal-protocol error type

gferon:
Goes together with https://github.com/Michael-F-Bryan/libsignal-protocol-rs/pull/70


rubdos:
Confirmed that this builds against Michael-F-Bryan/libsignal-protocol-rs#70

PR #16, opened 2020-09-28
Fixes to SMS, voice and secondary device registration

gferon:
This PR contains the initial changes I had to make to start working on [signal-bot-rs](https://github.com/gferon/signal-bot-rs). It looks like registration with SMS, voice and as secondary device works, but I can't be sure until I'm able to receive messages, hence why this PR is marked as draft.

PR #20, opened 2020-10-23
Finish registering / linking device

gferon:
Depends on https://github.com/Michael-F-Bryan/libsignal-protocol-rs/pull/70

* Fix / complete registration, confirm and link functions to be able to actually receive messages with what you obtain
* Implement HTTP error handling for the websocket
* Introduce new public `SignalingKey` type, mapping to `[u8; CIPHER_KEY_SIZE + MAC_KEY_SIZE]`
* Add new `pre_key` module containing the necessary API model structs to register pre-keys
* Comment out all resource paths, as you can't use them in `format!()` anyway
* Fix register and link examples

### DIFF
--- a/libsignal-service-actix/src/websocket.rs
+++ b/libsignal-service-actix/src/websocket.rs
@@ -1,6 +1,11 @@
 use actix::Arbiter;
 
-use awc::{error::WsProtocolError, ws, ws::Frame};
+use awc::{
+    error::{WsClientError, WsProtocolError},
+    http::StatusCode,
+    ws,
+    ws::Frame,
+};
 use bytes::Bytes;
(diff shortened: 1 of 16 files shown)


libsignal-service-actix/src/websocket.rs
```diff
@@ -23,7 +28,19 @@ pub enum AwcWebSocketError {
 
 impl From<AwcWebSocketError> for ServiceError {
     fn from(e: AwcWebSocketError) -> ServiceError {
-        todo!("error conversion {:?}", e)
```
rubdos, line 26: Heh thanks :)

rubdos:
#16 can be closed, then?

PR #18, opened 2020-10-18
MessageSender: sending messages

rubdos:

PR #23, opened 2020-10-27
Encrypt and send messages

gferon:
Supersedes https://github.com/Michael-F-Bryan/libsignal-service-rs/pull/18 since I can't push there.

TODO:
- [x] encrypt messages with existing sessions (requires receiving a message first)
- [x] delete old sessions for extra devices
- [x] establish new sessions with missing devices

Requires https://github.com/Michael-F-Bryan/libsignal-protocol-rs/pull/78 to be merged first.


libsignal-service/src/sender.rs
```diff
@@ -0,0 +1,151 @@
...
+    // Equivalent with `getEncryptedMessages`
+    fn create_encrypted_messages(
+        &mut self,
+        recipient: &crate::ServiceAddress,
+        unidentified_access: Option<bool>,
+        content: &[u8],
+    ) -> Result<Vec<OutgoingPushMessage>, MessageSenderError> {
+        let mut messages = vec![];
+
+        let myself = *recipient == self.cipher.local;
```
rubdos, line 104: I made `ServiceAddress::matches` for this, although it seems like you dropped that commit (3d1c0238f26731c456b8077399fbfb021eff4087, you should be able to fetch it from #18). `matches` is more fuzzy than `Eq`: `Eq` takes into account that both `e164` and `uuid` needs to be the exact same, but with `matches` only one of them has to match.

libsignal-service/src/sender.rs
```diff
@@ -0,0 +1,151 @@
...
+            let message = self.create_encrypted_message(
+                recipient,
+                unidentified_access,
+                DEFAULT_DEVICE_ID,
+                content,
+            )?;
+            messages.push(message);
+        }
+
+        // TODO: get sub-device-sessions here to send to myself!
```
rubdos, line 118: I think I'd still like this in this PR; the way I debug is with Signal Desktop, which is sending to myself :-)

libsignal-service/src/cipher.rs
```diff
@@ -177,6 +182,63 @@ impl ServiceCipher {
...
+                    message_part_count += 1;
+                }
+
+                message_part_count * 160
+            };
+
+            let session_cipher =
+                SessionCipher::new(&self.context, &self.store, address)?;
+
+            let _session_version = session_cipher.get_session_version()?;
```
rubdos, line 209: Unused?

rubdos: If we're only supporting one single version, maybe error out on the wrong version.

libsignal-service/src/cipher.rs
```diff
@@ -2,22 +2,27 @@ use crate::{
...
 use prost::Message;
 
 /// Decrypts incoming messages and encrypts outgoing messages.
 ///
 /// Equivalent of SignalServiceCipher in Java.
+#[derive(Clone, Debug)]
 pub struct ServiceCipher {
     store: StoreContext,
-    local: ServiceAddress,
+    pub local: ServiceAddress,
```
rubdos, line 25: `pub(crate)` seems more appropriate. I had personally changed the constructor of `MessageSender` to have another copy of the local address, but either way is okay.

rubdos (approved):
Some nitpicks, one feature request, looks good otherwise!

libsignal-service/src/cipher.rs
```diff
@@ -177,6 +182,63 @@ impl ServiceCipher {
...
+            let get_padded_message_length = |len: usize| {
+                let message_length_with_terminator = len + 1;
+                let mut message_part_count =
+                    message_length_with_terminator / 160;
+                if message_length_with_terminator % 160 != 0 {
+                    message_part_count += 1;
+                }
+
+                message_part_count * 160
+            };
```
rubdos, line 204: Maybe we should use [`block-padding`'s `Iso7816`](https://docs.rs/block-padding/0.2.1/block_padding/enum.Iso7816.html) for this, and replace the one in `decrypt` too.

rubdos (changes requested):

libsignal-service/src/sender.rs
gferon: Hmm, I might be wrong but I don't see `ServiceAddress::matches` anywhere here: https://github.com/Michael-F-Bryan/libsignal-service-rs/pull/18/files

libsignal-service/src/sender.rs
gferon: Just added it, it works now! I've tested with "Note to self".

libsignal-service/src/push_service.rs
```diff
@@ -205,28 +229,24 @@ pub trait PushService {
     type WebSocket: WebSocketService;
     type ByteStream: futures::io::AsyncRead + Unpin;
 
-    async fn get<T>(&mut self, path: &str) -> Result<T, ServiceError>
+    async fn get<T>(&self, path: &str) -> Result<T, ServiceError>
```
rubdos, line 232: Actually, for the trait, I think it's still useful to use `&mut`, because I'm not quite sure that all http backends will have internal mutability for writing to a socket. What do you think?

libsignal-service/src/utils.rs
```diff
@@ -58,7 +57,7 @@ pub mod serde_optional_base64 {
 }
 
 pub mod serde_public_key {
-    use libsignal_protocol::keys::PublicKey;
+    use libsignal_protocol::{keys::PublicKey, Base64};
```
rubdos, line 60: This import is dead afaik.

libsignal-service/src/push_service.rs
gferon: `hyper` doesn't need interior mutability as well, so I don't think it's a concern (see: https://docs.rs/hyper/0.13.8/hyper/client/struct.Client.html#method.get).

libsignal-service/src/cipher.rs
```diff
@@ -184,27 +235,51 @@ struct Plaintext {
...
 }
 
+fn add_padding(version: u32, contents: &[u8]) -> Result<Vec<u8>, ServiceError> {
+    if version < 2 {
+        Err(ServiceError::InvalidFrameError {
+            reason: format!("Unknown version {}", version),
+        })
+    } else if version == 2 {
+        Ok(contents.to_vec())
+    } else {
```
rubdos, line 245: Is this how it's padded in Java? Seems weird that *>2 goes the same way, no?

rubdos: Hmm, I notice it's the same in `strip_padding`, which I translated from Java...

rubdos (commented):
Some minor questions that I still have.

libsignal-service/src/cipher.rs
gferon: Yes, I also translated this from Java :smile:

rubdos (approved):

PR #27, opened 2020-10-29
Sealed session cipher

### LINKED ISSUE #10: Decrypt UnidentifiedSender
Envelopes with `r#type() == UnidentifiedSender` currently panic: "UnidentifiedSender requires SealedSessionCipher, please report a bug against libsignal-service-rs."

This needs to be handled, such as [in Java's SignalServiceCipher](https://github.com/signalapp/libsignal-service-java/blob/1a01c22636ad2c4e24e911ed7c4eb95f14cc58d6/java/src/main/java/org/whispersystems/signalservice/api/crypto/SignalServiceCipher.java#L250)

gferon:
Delivery & read receipts are delivered (at least by the Signal Android app) as unidentified sender. This requires to implement the [SealedSessionCipher](https://github.com/signalapp/libsignal-metadata-java/blob/master/java/src/main/java/org/signal/libsignal/metadata/SealedSessionCipher.java) from `libsignal-service-java`.

- [x] Decrypt messages from unidentified senders
- [x] Encrypt messages as a unidentified sender

Requires https://github.com/Michael-F-Bryan/libsignal-protocol-rs/pull/79
Fixes #10


gferon:
@rubdos I think you can already review this part, I'll add `encrypt` in this PR later, but I don't think it'll change much.

libsignal-service/src/cipher.rs
```diff
@@ -83,84 +97,99 @@ impl ServiceCipher {
...
+                        &PreKeySignalMessage::deserialize(
+                            &self.context,
+                            ciphertext,
+                        )?,
+                    )?
+                    .as_slice()
+                    .to_vec();
+                let version =
+                    self.store_context.load_session(&sender)?.state().version();
+                strip_padding(version, &mut data)?;
```
rubdos, line 130: Starting to think that `strip_padding` should accept `&'a [u8]` and return `&'a [u8]`... Then we don't need the mutability. Maybe another time...

libsignal-service/src/cipher.rs
```diff
@@ -83,84 +97,99 @@ impl ServiceCipher {
...
-            }
-        };
-
-        let mut plaintext = match envelope.r#type() {
+        let plaintext = match envelope.r#type() {
             Type::PrekeyBundle => {
+                let sender = self.get_preferred_protocol_address(
+                    envelope.source_address().expect("Envelope with source"),
+                    envelope.source_device(),
+                )?;
```
rubdos, line 105: Duplicate code with `Type::Ciphertext`. Does it make sense to put this before the match? I guess the compiler is smart enough to only call it on those two types.

rubdos: Same with metadata afaict

libsignal-service/src/cipher.rs
```diff
@@ -83,84 +97,99 @@ impl ServiceCipher {
...
+                    sender_uuid,
+                    sender_e164,
+                    device_id,
+                    padded_message: mut data,
+                    version,
+                } = self
+                    .sealed_session_cipher
+                    .decrypt(ciphertext, envelope.timestamp())?;
+                let sender = ServiceAddress {
+                    e164: sender_e164.unwrap_or("".into()), // XXX: FIXME
```
rubdos, line 173: XXX -> this means e164 should become `Option<>`, already? Whisperfish currently always assumes e164, because we're old skool.

libsignal-service/src/cipher.rs
```diff
@@ -83,84 +97,99 @@ impl ServiceCipher {
...
+                    needs_receipt: false,
+                };
+                strip_padding(version, &mut data)?;
+                Plaintext { metadata, data }
             }
             _ => {
-                unreachable!("conditions checked in previous match");
+                // else
+                return Err(ServiceError::InvalidFrameError {
+                    reason: "Envelope has unknown type.".into(),
```
rubdos, line 189:
```suggestion
                    reason: format!("Envelope has unknown type {:?}.", envelope.r#type()),
```

libsignal-service/src/cipher.rs
rubdos: Maybe `strip_padding` calls should go into the decrypt functions themselves? I think that would make sense.

libsignal-service/src/sealed_session_cipher.rs
```diff
@@ -0,0 +1,533 @@
...
+    sender_device_id: i32,
+    sender_uuid: Option<String>,
+    sender_e164: Option<String>,
+    expiration: u64,
+    pub certificate: Vec<u8>,
+    pub signature: Vec<u8>,
+}
+
+impl SenderCertificate {
+    // XXX: Result
```
rubdos, line 87: XXX

libsignal-service/src/sealed_session_cipher.rs
```diff
@@ -0,0 +1,533 @@
...
+                Ok(Self {
+                    r#type: match message::Type::from_i32(message_type) {
+                        Some(message::Type::Message) => CiphertextType::Signal,
+                        Some(message::Type::PrekeyMessage) => {
+                            CiphertextType::PreKey
+                        }
+                        _ => {
+                            return Err(
+                                SealedSessionError::InvalidMetadataMessageError(
+                                    "Wrong message type".into(),
```
rubdos, line 403:
```suggestion
                                    format!("Wrong message type ({})", message::Type::from_i32(message_type)),
```

rubdos (changes requested):
On top of the comments I made, can we import some unit tests and maybe integration tests from the Java version? 500LOC is a ton, especially in terms of pure crypto-protocol code. We really ought to have some known-answer tests here.

Otherwise looks *really* clean, good job.

libsignal-service/src/cipher.rs
gferon: Yes, I initially wanted to change it. It'll be a slightly annoying breaking change though. Maybe we can defer it for later? (I can open an issue).

libsignal-service/src/cipher.rs
gferon: I only put it there to be consistent with `libsignal-metadata-java` which returns the padded message, but it doesn't really make any sense if we keep the `SealedSessionCipher` in this crate.

If we split the crates though, it might.

gferon:
@rubdos do you think I should move this into a separate `libsignal-metadata-rs` crate?

rubdos:
> @rubdos do you think I should move this into a separate `libsignal-metadata-rs` crate?

I don't think it's necessary, as discussed on Matrix. Your pick, if you want to move it. Just write out your argumentation in this PR for future reference :-)

libsignal-service/src/cipher.rs
gferon: Yeah, I moved it for better readability, I didn't really like the double `match` from before with the second one having a catch all saying `unreachable!("we checked that before")`. Let me see if I come up with something more elegant.

libsignal-service/src/cipher.rs
```diff
@@ -212,10 +232,7 @@ impl ServiceCipher {
                 session_cipher.get_remote_registration_id()?;
             let body = base64::encode(message.serialize()?);
             use crate::proto::envelope::Type;
-            let message_type = match message
-                .get_type()
-                .map_err(libsignal_protocol::Error::InternalError)?
-            {
+            let message_type = match message.get_type()? {
```
rubdos, line 235: I've put this change in master already, because it's in `protocol`'s master.

libsignal-service/src/cipher.rs
gferon: Ok, for now I would not change it just because of separation of concern, I think the `decrypt` functions should return the stream as it's intended to be.

gferon:
@rubdos I think this is ready for the final review!

libsignal-service/src/cipher.rs
```diff
@@ -19,23 +21,31 @@ use prost::Message;
...
+    sealed_session_cipher: SealedSessionCipher,
 }
 
 impl ServiceCipher {
     pub fn from_context(
         context: Context,
-        local_address: ServiceAddress,
         store_context: StoreContext,
+        local_address: ServiceAddress,
+        certificate_validator: CertificateValidator,
```
rubdos, line 37: Would it make sense to construct the `CertificateValidator` here, based on `context`?

libsignal-service/src/cipher.rs
rubdos: You figured out this yet?

libsignal-service/src/cipher.rs
rubdos: Postponed to #46

libsignal-service/src/content.rs
```diff
@@ -2,17 +2,16 @@ pub use crate::{
     proto::{
         attachment_pointer::Flags as AttachmentPointerFlags,
         data_message::Flags as DataMessageFlags, data_message::Reaction,
-        group_context::Type as GroupType, sync_message, AttachmentPointer,
-        CallMessage, DataMessage, GroupContext, ReceiptMessage, SyncMessage,
-        TypingMessage,
+        group_context::Type as GroupType, AttachmentPointer, CallMessage,
+        DataMessage, ReceiptMessage, SyncMessage, TypingMessage,
```
rubdos, line 6: The structs in the [`sync_message` module](`https://rubdos.gitlab.io/whisperfish/doc/libsignal_service/proto/sync_message/index.html`) are necessary in client implementation. That's why I added that to the `pub use`.

```sh
whisperfish
❯ rg sync_message
src/worker/client.rs
590:                use libsignal_service::content::sync_message;
593:                    sent: Some(sync_message::Sent {
```

libsignal-service/src/sealed_session_cipher.rs
```diff
@@ -0,0 +1,1070 @@
...
+    ) -> Self {
+        Self {
+            context,
+            store_context,
+            local_address,
+            certificate_validator,
+        }
+    }
+
+    #[allow(dead_code)]
```
rubdos, line 214: Why is that? Maybe add a comment explaining, if still needed.

libsignal-service/src/sealed_session_cipher.rs
```diff
@@ -0,0 +1,1070 @@
...
+        &self,
+        certificate: &SenderCertificate,
+    ) -> Result<Address, SealedSessionError> {
+        if let Some(ref sender_uuid) = certificate.sender_uuid {
+            let address =
+                Address::new(sender_uuid, certificate.sender_device_id as i32);
+            if self.store_context.contains_session(&address)? {
+                return Ok(address);
+            }
+        } else if let Some(ref sender_e164) = certificate.sender_e164 {
```
rubdos, line 445: `else` shouldn't go here.  The logic is the same as in the other cipher, can this be somehow refactored?

libsignal-service/src/sealed_session_cipher.rs
```diff
@@ -0,0 +1,1070 @@
...
+            CiphertextType::PreKey => {
+                let msg = session_cipher.decrypt_pre_key_message(
+                    &PreKeySignalMessage::deserialize(
+                        &self.context,
+                        &message.content,
+                    )?,
+                )?;
+                msg.as_slice().to_vec()
+            }
+            _ => unreachable!("unknown message from unidentified sender type"),
```
rubdos, line 485: Shouldn't this  be an error instead of a panic, to be forward-compatible?

libsignal-service/src/sealed_session_cipher.rs
```diff
@@ -0,0 +1,1070 @@
...
+                    }),
+                    _ => Err(SealedSessionError::InvalidCertificate),
+                }
+            }
+            _ => Err(SealedSessionError::InvalidCertificate),
+        }
+    }
+}
+
+// impl TryInto<Vec<u8>> for ServerCertificate {
```
rubdos, line 640: Commented code, should this go, or stay, or be rewritten, or be put in an issue?

rubdos (approved):
One bug spotted, everything else are general remarks.

libsignal-service/src/sealed_session_cipher.rs
gferon: Oooh, because I haven't implemented _sending_ of encrypted messages in the `MessageSender` yet.

rubdos: Should that still go in this PR, or should that be an issue? If the latter, I suggest writing a comment on the `allow(dead_code)` linking to the issue.

libsignal-service/src/sealed_session_cipher.rs
gferon: well, it has been tested in the caller, it is an assert - it means we really mis-used the API in this case, so I'd still panic.

libsignal-service/src/sealed_session_cipher.rs
gferon: bug fixed, thanks! I shared the implementation of `get_preferred_protocol_address` with the `SessionCipher`.

libsignal-service/src/cipher.rs
gferon: well, the CertificateValidator needs a trusted root certificate which comes from ServiceConfiguration. I'm fine passing this as a parameter instead, as it's already done for AwcPushService.

libsignal-service/src/cipher.rs
gferon: Not worth the extra effort, as the branch from `UnidentifiedSender` is using a `ProtocolAddress` so you cannot easily have a match arm before to return `(sender, metadata)` without un-necessarily converting types.

rubdos (approved):

PR #32, opened 2020-11-03
Switch to using GitHub Workflows for CI

gferon:


rubdos:
@gferon you think we can merge this?

PR #33, opened 2020-11-03
Enable GitHub workflows

rubdos:
Overrules #32, because I don't think Gabriel has the right to change GH Workflows


rubdos:
Nope, I guess it needs to be in master branch.

gferon:
No, it ran! But there's a syntax error.

Issue #34, opened 2020-11-03
AttachmentV3


rubdos:
Attachments come in two versions, and I am implementing V2. V3 is there, but is slightly more involved, and will only become interesting for me in Whisperfish 1.0. Leaving this as a note.

PR #35, opened 2020-11-04
Add attachment upload code

### LINKED ISSUE #26: Send attachments

rubdos:
Fixes #26

### DIFF
--- a/libsignal-service-actix/Cargo.toml
+++ b/libsignal-service-actix/Cargo.toml
@@ -12,7 +12,9 @@ libsignal-protocol = { git = "https://github.com/Michael-F-Bryan/libsignal-proto
 
 awc = { version = "2.0.0", features=["rustls"] }
 actix = "0.10.0"
+actix-http = "2.1.0"
 actix-rt = "1.1"
+actix-multipart-rfc7578 = {version = "0.3.0-rc", git="https://github.com/rubdos/rust-multipart-rfc7578.git", branch="actix-2"} # Not an "official" actix project
 serde_json = "1.0"
 futures = "0.3"
 bytes = "0.5"
--- a/libsignal-service-actix/Cargo.toml
+++ b/libsignal-service-actix/Cargo.toml
@@ -14,7 +14,7 @@ awc = { version = "2.0.0", features=["rustls"] }
 actix = "0.10.0"
 actix-http = "2.1.0"
 actix-rt = "1.1"
-actix-multipart-rfc7578 = {version = "0.3.0-rc", git="https://github.com/rubdos/rust-multipart-rfc7578.git", branch="actix-2"} # Not an "official" actix project
+mpart-async = "0.4"
 serde_json = "1.0"
 futures = "0.3"
 bytes = "0.5"
--- a/libsignal-service-actix/Cargo.toml
+++ b/libsignal-service-actix/Cargo.toml
@@ -22,6 +22,7 @@ rustls = "0.18"
 url = "2.1"
 serde = "1.0"
 log = "0.4.8"
+rand = "0.7"
 
 failure = "0.1.5"
 thiserror = "1.0"
--- a/libsignal-service-actix/src/push_service.rs
+++ b/libsignal-service-actix/src/push_service.rs
@@ -1,5 +1,6 @@
 use std::{sync::Arc, time::Duration};
 
+use actix_multipart_rfc7578::client::multipart;
 use awc::{
     error::PayloadError, http::StatusCode, Client, ClientResponse, Connector,
 };
@@ -178,6 +179,70 @@ impl PushService for AwcPushService {
                 })
                 .into_async_read(),
         ))
+    }
+
+    async fn post_to_cdn0<'s, C: std::io::Read + Send + 's>(
+        &mut self,
+        path: &str,
+        value: &[(&str, &str)],
+        file: Option<(&str, &'s mut C)>,
+    ) -> Result<(), ServiceError> {
+        let url = Url::parse(&self.cfg.cdn_urls[&0])
+            .expect("valid cdn base url")
+            .join(path)
+            .expect("valid CDN path");
+
+        log::debug!("AwcPushService::post_to_cdn({:?})", url);
+        let mut client = self.client.post(url.as_str());
+
+        let mut form = multipart::Form::default();
+
+        for (k, v) in value {
+            form.add_text(*k, *v);
+        }
+
+        if let Some((filename, file)) = file {
+            // XXX Actix doesn't cope with none-'static lifetimes
+            // https://docs.rs/actix-web/3.2.0/actix_web/body/enum.Body.html
+            let mut buf = Vec::new();
+            file.read_to_end(&mut buf)
+                .expect("infallible Read instance");
+            form.add_reader(filename, std::io::Cursor::new(buf));
+            client = client.header("Content-Type", "application/octet-stream");
+        }
+
+        let content_type = form.content_type();
+
+        // XXX Amazon S3 needs the Content-Length, but we don't know it without depleting the whole
+        // stream.  Sadly, Content-Length != contents.len(), but should include the whole form.
+        let mut body = multipart::Body::from(form);
+        let mut body_contents = vec![];
+        use actix_http::body::MessageBody;
+        use futures::stream::StreamExt;
+        let mut stream = futures::stream::poll_fn(move |ctx| {
+            let body = std::pin::Pin::new(&mut body);
+            MessageBody::poll_next(body, ctx)
+        });
+        while let Some(b) = stream.next().await {
+            body_contents.extend(b.map_err(|e| ServiceError::SendError {
+                reason: e.to_string(),
+            })?);
+        }
+
+        let mut response = client
+            .content_type(content_type)
+            .content_length(body_contents.len() as u64)
+            .send_body(body_contents)
+            .await
+            .map_err(|e| ServiceError::SendError {
+                reason: e.to_string(),
+            })?;
+
+        log::debug!("AwcPushService::put response: {:?}", response);
+
+        Self::from_response(&mut response).await?;
+
+        Ok(())
     }
 
     async fn ws(
--- a/libsignal-service-actix/src/push_service.rs
+++ b/libsignal-service-actix/src/push_service.rs
@@ -183,7 +183,7 @@ impl PushService for AwcPushService {
     async fn post_to_cdn0<'s, C: std::io::Read + Send + 's>(
         &mut self,
         path: &str,
-        value: &[(&str, &str)],
+        mut value: &[(&str, &str)],
         file: Option<(&str, &'s mut C)>,
     ) -> Result<(), ServiceError> {
         let url = Url::parse(&self.cfg.cdn_urls[&0])
@@ -196,8 +196,23 @@ impl PushService for AwcPushService {
 
         let mut form = mpart_async::client::MultipartRequest::default();
 
-        for &(k, v) in value {
+        // XXX: mpart-async has a *very* peculiar ordering of the form items,
+        //      and Amazon S3 expects them in a very specific order (i.e., the file contents should
+        //      go last.
+        //
+        //      Currently, talking version 4.1, mpart-async internally stores a STATE and a list of
+        //      FIELDS.  The very first item to be added is set as STATE, while every next items is
+        //      appended to the vector of FIELDS.
+        //      When yielding as a stream, msync-async returns the current STATE, and then uses
+        //      `Vec::pop` to decide on the next state.
+        //      This, however, is an implementation detail to us, which we are exploiting here.
+        //
+        //      https://github.com/cetra3/mpart-async/issues/16
+        if !value.is_empty() {
+            let (k, v) = value[0];
             form.add_field(k, v);
+
+            value = &value[1..];
         }
 
         if let Some((filename, file)) = file {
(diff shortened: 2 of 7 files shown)


libsignal-service-actix/Cargo.toml
```diff
@@ -12,7 +12,9 @@ libsignal-protocol = { git = "https://github.com/Michael-F-Bryan/libsignal-proto
 
 awc = { version = "2.0.0", features=["rustls"] }
 actix = "0.10.0"
+actix-http = "2.1.0"
 actix-rt = "1.1"
+actix-multipart-rfc7578 = {version = "0.3.0-rc", git="https://github.com/rubdos/rust-multipart-rfc7578.git", branch="actix-2"} # Not an "official" actix project
```
rubdos, line 17: Tracking PR https://github.com/ferristseng/rust-multipart-rfc7578/pull/19

libsignal-service-actix/Cargo.toml
```diff
@@ -12,7 +12,9 @@ libsignal-protocol = { git = "https://github.com/Michael-F-Bryan/libsignal-proto
 
 awc = { version = "2.0.0", features=["rustls"] }
 actix = "0.10.0"
+actix-http = "2.1.0"
```
rubdos, line 15: Is this used?

libsignal-service/Cargo.toml
```diff
@@ -12,6 +12,7 @@ async-trait = "0.1.30"
 url = { version = "2.1.1", features = ["serde"] }
 base64 = "0.13"
 bytes = "0.5"
+chrono = "0.4"
```
rubdos, line 15: It's a bit sad that we include `chrono` for just `Utc::now().timestamp_millis()`.

libsignal-service/src/sender.rs
```diff
@@ -70,6 +100,67 @@ where
...
+        let (key, iv) = {
+            use rand::RngCore;
+            let mut key = [0u8; 64];
+            let mut iv = [0u8; 16];
+            // thread_rng is guaranteed to be cryptographically secure
+            rand::thread_rng().fill_bytes(&mut key);
+            rand::thread_rng().fill_bytes(&mut iv);
+            (key, iv)
+        };
+        // XXX upstream Java seems to pad this further
```
rubdos, line 122: We may want to fix this in this PR

libsignal-service-actix/Cargo.toml
rubdos: It is in the latest push!

libsignal-service-actix/src/push_service.rs
```diff
@@ -182,6 +183,70 @@ impl PushService for AwcPushService {
...
     }
 
+    async fn post_to_cdn0<'s, C: std::io::Read + Send + 's>(
+        &mut self,
+        path: &str,
+        value: &[(&str, &str)],
+        file: Option<(&str, &'s mut C)>,
+    ) -> Result<(), ServiceError> {
+        let url = Url::parse(&self.cfg.cdn_urls[&0])
+            .expect("valid cdn base url")
```
gferon, line 193: I guess now that we use `thiserror` everywhere, it would be nicer to have the `uri::ParseError` mapped there instead of unwrapping?

libsignal-service-actix/src/push_service.rs
```diff
@@ -182,6 +183,70 @@ impl PushService for AwcPushService {
...
+            file.read_to_end(&mut buf)
+                .expect("infallible Read instance");
+            form.add_reader(filename, std::io::Cursor::new(buf));
+            client = client.header("Content-Type", "application/octet-stream");
+        }
+
+        let content_type = form.content_type();
+
+        // XXX Amazon S3 needs the Content-Length, but we don't know it without depleting the whole
+        // stream.  Sadly, Content-Length != contents.len(), but should include the whole form.
```
gferon, line 219: Just being curious here: how's `Content-Length` is not `contents.len()`? There _must_ be something we can do to calculate the right value... but I guess you already tried that :smile:

libsignal-service/src/sender.rs
```diff
@@ -70,6 +100,67 @@ where
...
+                    AttachmentPointerFlags::Borderless as u32
+                } else {
+                    0
+                },
+            ),
+            width: Some(spec.width),
+            height: Some(spec.height),
+            caption: spec.caption,
+            blur_hash: spec.blur_hash,
+            upload_timestamp: Some(chrono::Utc::now().timestamp_millis() as u64),
```
gferon, line 157:
```suggestion
            upload_timestamp: Some(chrono::Utc::now().timestamp_millis() as u64),
            upload_timestamp: Some(SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).as_millis() as u64)
```

libsignal-service/Cargo.toml
gferon: I added a suggestion below to not have to import `chrono`.

libsignal-service-actix/src/push_service.rs
```diff
@@ -182,6 +183,70 @@ impl PushService for AwcPushService {
...
+        let mut body = multipart::Body::from(form);
+        let mut body_contents = vec![];
+        use actix_http::body::MessageBody;
+        use futures::stream::StreamExt;
+        let mut stream = futures::stream::poll_fn(move |ctx| {
+            let body = std::pin::Pin::new(&mut body);
```
gferon, line 225: maybe `pin_mut!(body)` instead?

libsignal-service/src/push_service.rs
```diff
@@ -345,6 +388,45 @@ pub trait PushService {
...
+            ("key", &attrs.key),
+            ("policy", &attrs.policy),
+            ("Content-Type", "application/octet-stream"),
+            ("x-amz-algorithm", &attrs.algorithm),
+            ("x-amz-credential", &attrs.credential),
+            ("x-amz-date", &attrs.date),
+            ("x-amz-signature", &attrs.signature),
+        ];
+
+        let mut digester = crate::digeststream::DigestingReader::new(content);
```
gferon, line 419: In the end, if this is just a SHA-256 hasher, why not use it directly (and return the hash from `self.post_to_cdn0`)? I suppose there's a reason that I missed here.

gferon (commented):
Some minor comments, but it looks good to me! Looking forward to testing it :test_tube:

libsignal-service-actix/src/push_service.rs
rubdos: Should we not test this in the constructor instead?

libsignal-service-actix/src/push_service.rs
rubdos: Because `contents` contains the attachment, and what you send to the Amazon S3 server is a form (yes, an HTML form), that contains a bunch of field *and* the attachment. As such, it depends on how the serialization of the form happens, which is why I serialize before (to memory), and count, and then send.

AttachmentV3 does not have that issue, by the way, but is way more involved (has a few extra requests). It has as advantage that you can stream, without knowing contents-length, and resume an upload later on.

libsignal-service/src/push_service.rs
rubdos: Because I'd like all interfaces to be streaming instead of memory-committed. `post_to_cdn0` *probably* cannot be streaming though, so you comment makes sense in that regard.

Otoh, I'd keep the `DigestingReader` because attachment V3 supposedly can be streamed all the way.

shekohex:
I'm just curious, as I'm implementing something similar in my E2EE Chat (https://github.com/owlchat)

How do you think you would handle the partial upload (in the client for now)? I know it's maybe out of the scope of the current PR .. do you want to discuss this further?😊

rubdos:
> How do you think you would handle the partial upload (in the client for now)? I know it's maybe out of the scope of the current PR .. do you want to discuss this further?blush

Do you mean with the current code, or with AttachmentsV3? In the current code, there's simply no partial upload. My use case is Whisperfish, which simply doesn't have the architecture (yet) to do partial uploads. I do however want to have partial uploads (through V3) in the future! Does that answer your question?

Then there is of course the API that needs to be exposed to upload files for V3... That maybe needs some thorough discussion :-)

shekohex:
Yes, I was referring to Attachment V3 and how signal client do partial uploads.
Anyway, we should discuss more over discord for example? My Discord username is the same as my Github :)

rubdos:
> Yes, I was referring to Attachment V3 and how signal client do partial uploads.

Aha!

> Anyway, we should discuss more over discord for example? My Discord username is the same as my Github :)

We're talking about developing a secure and private messenger here, and we're gonna do the talking over Discord? I'd find that ironic hehe.  We're mostly on Matrix (@rubdos:rubdos.be) and IRC, but I would actually suggesting discussing the V3 interface in the relevant issue #34, to have a public track record of it.

libsignal-service-actix/src/push_service.rs
gferon: Puh. Ok, I accept this (foul) explanation :smiley:

libsignal-service-actix/src/push_service.rs
gferon: Of `ServiceConfiguration`? Well, yes, but... catch all errors and never rely on the implementation of another struct :crab:

libsignal-service-actix/Cargo.toml
rubdos: @gferon this crate does seem kind-of abandoned. Would you have any suggestion how to resolve this? Maybe contribute it to Actix directly?

rubdos: Actix is tracking that here: https://github.com/actix/actix-web/issues/195

So I guess I'll take a crack at that to get this merged.

libsignal-service-actix/src/push_service.rs
rubdos: Something very weird happens if I try:

```
error[E0507]: cannot move out of `body`, a captured variable in an `FnMut` closure
   --> libsignal-service-actix/src/push_service.rs:225:13
    |
220 |         let mut body = multipart::Body::from(form);
    |             -------- captured outer variable
...
225 |             futures::pin_mut!(body);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^
    |             |
    |             move occurs because `body` has type `actix_multipart_rfc7578::client::multipart::Body<'_>`, which does not implement the `Copy` trait
    |             help: consider borrowing here: `&$x`
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

warning: variable does not need to be mutable
   --> libsignal-service-actix/src/push_service.rs:220:13
    |
220 |         let mut body = multipart::Body::from(form);
    |             ----^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default
```

gferon: You can ignore my comment then, thanks for trying.

libsignal-service-actix/Cargo.toml
rubdos: That issue is quite some work... I gave it a shot, then noticed that one of the Actix devs has the `mpart-async` crate that solves this a lot cleaner. I've rewritten it with that, now I'll clean that up a bit.

libsignal-service-actix/src/push_service.rs
rubdos: -> https://github.com/Michael-F-Bryan/libsignal-service-rs/issues/48

libsignal-service-actix/Cargo.toml
rubdos: Some extra comments:

- Actix basically expects `Stream<Item=Bytes>` for its chunks
- I think `tokio::io::AsyncRead` would be the "logical" block
- Converting `tokio::io::AsyncRead` into `Stream<Item=Bytes>` is [non-trivial](https://github.com/rust-lang/futures-rs/issues/2006).
- The current `std::io::Read` is *not good*. It blocks. That's terrible. We're allocating, that's slightly less terrible, but still bad.

I'd like to merge this this weekend, @gferon, you got any suggestions how to clean this up? Just pass on the whole `Vec<u8>` for the `cdn0` implementation, and when we start #34, do it decently (because that's an actual streaming API)? The `cdn0` version needs to know `Content-Length` anyway, so we might as well keep this kinda dirty.

libsignal-service/src/sender.rs
rubdos: Fixed in the last commit. @gferon would you mind taking a second look that I copied that Math thing correctly from Java?

gferon (approved):
Awesome! I suppose you have tested that it works :mailbox_with_mail:

rubdos:
I've filed https://github.com/cetra3/mpart-async/issues/16 and added that link to the `XXX` comment of 45fd53e

rubdos:
> Awesome! I suppose you have tested that it works mailbox_with_mail

And now it is tested!

image

libsignal-service-actix/Cargo.toml
gferon: > Just pass on the whole `Vec<u8>` for the `cdn0` implementation, and when we start #34, do it decently (because that's an actual streaming API)? The `cdn0` version needs to know `Content-Length` anyway, so we might as well keep this kinda dirty.

This definitely sounds sensible, either `Vec<u8>` or `&[u8]` if the caller should retain ownership of the data somehow.

gferon (approved):
:rocket:

PR #39, opened 2020-11-06
Make e164 optional in ServiceAddress

gferon:
This is a necessary step for future improvements, like #27.

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -139,12 +139,14 @@ where
                             ),
                         )?;
                     }
-                    self.cipher.store_context.delete_session(
-                        &libsignal_protocol::Address::new(
-                            &recipient.e164,
-                            *extra_device_id,
-                        ),
-                    )?;
+                    if let Some(ref e164) = recipient.e164 {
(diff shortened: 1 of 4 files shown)


libsignal-service/src/sender.rs
```diff
@@ -139,12 +139,14 @@ where
                             ),
                         )?;
                     }
-                    self.cipher.store_context.delete_session(
-                        &libsignal_protocol::Address::new(
-                            &recipient.e164,
-                            *extra_device_id,
-                        ),
-                    )?;
+                    if let Some(ref e164) = recipient.e164 {
```
rubdos, line 142: Sessions with Uuid should be deleted too, no? Maybe `get_preferred_protocol_address`?

rubdos (changes requested):

libsignal-service/src/sender.rs
gferon: Yes, it was already done before? (open the code above).

rubdos: Yep, missed that, sorry for that.

rubdos (approved):
LGTM

PR #40, opened 2020-11-07
AccountManager::update_pre_key_bundle

rubdos:
I was looking at https://github.com/gferon/signal-bot-rs/blob/wip/src/manager.rs#L361 and figured this should be common code (as it is in the Java code).

I have integrated this in Whisperfish, but haven't tested yet. Will do that tomorrow.

Good night! ~R

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -67,7 +67,16 @@ impl<Service: PushService> AccountManager<Service> {
         pre_keys_offset_id: u32,
         next_signed_pre_key_id: u32,
     ) -> Result<(u32, u32), ServiceError> {
-        let prekey_count = self.service.get_pre_key_status().await?.count;
+        let prekey_count = match self.service.get_pre_key_status().await {
+            Ok(status) => status.count,
+            Err(ServiceError::Unauthorized) => {
+                log::info!("Got Unauthorized when fetching pre-key status. Assuming first installment.");
+                // Additionally, the second PUT request will fail if this really comes down to an
+                // authorization failure.
+                0
+            }
+            Err(e) => return Err(e),
+        };
         log::trace!("Remaining pre-keys on server: {}", prekey_count);
 
         if prekey_count >= PRE_KEY_MINIMUM {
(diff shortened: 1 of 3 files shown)


libsignal-service/src/account_manager.rs
```diff
@@ -1,15 +1,25 @@
...
 }
 
+const PRE_KEY_MINIMUM: u32 = 10;
+const PRE_KEY_BATCH_SIZE: u32 = 100;
+
 impl<Service: PushService> AccountManager<Service> {
-    pub fn new(service: Service) -> Self {
-        Self { service }
+    pub fn new(context: Context, service: Service) -> Self {
+        Self { service, context }
```
gferon, line 22: being pedantic here, but maybe swap the order of the fields to match the constructor?

gferon (approved):
Awesome! I only have one minor nitpick. It's really nice to integrate more common code now that we start having different consumers of the crate.

gferon (approved):
Just fixed the build of the example, will merge when CI is green :rocket:

libsignal-service/src/account_manager.rs
```diff
@@ -42,4 +52,57 @@ impl<Service: PushService> AccountManager<Service> {
...
+    /// Checks the availability of pre-keys, and updates them as necessary.
+    ///
+    /// Parameters are the protocol's `StoreContext`, and the offsets for the next pre-key and
+    /// signed pre-keys.
+    ///
+    /// Equivalent to Java's RefreshPreKeysJob
+    pub async fn update_pre_key_bundle(
+        &mut self,
+        store_context: StoreContext,
+        pre_keys_offset_id: u32,
```
gferon, line 65: What about outputting the new value (so the caller doesn't have to know how many pre-keys we generate, if we did that etc.) as either a return value, or taking the parameter as `&mut pre_keys_offset_id`?

rubdos: Good idea. Whisperfish (pre 0.5) used a random u32 as next pre-key ID, and I just look at those that I have stored and go from there. But returning it is very useful indeed.

libsignal-service/src/account_manager.rs
rubdos: I was wondering whether the `StoreContext` from below should maybe also go in the constructor.

PR #43, opened 2020-11-08
Fix (or ignore) various clippy warnings

gferon:


rubdos:
Should we disallow clippy failure, while we're at it?

PR #45, opened 2020-11-09
Replace Travis badge by docs link

gferon:


rubdos:
@thibautvdv seems like the `index.html` part was unneeded, because it just works now!

PR #51, opened 2020-11-17
Add default-off feature `prefer-e164` for legacy

rubdos:
Legacy applications (notably Whisperfish) are having a hard time™
migrating to UUID-identified sessions, and they would prefer keeping
E164-identified sessions for a while.

This feature flag returns a E164-identified session whenever possible.


gferon (approved):
Never break downstream users :wink:

PR #50, opened 2020-11-15
Strong types for ServiceAddress

rubdos:
- [x] Rename `e164` fields to `phonenumber`?

Fixes #47 and #67

#67 is unrelated, but since I passed over that code...

PR #52, opened 2020-11-21
Contact discovery

rubdos:
Small patch, but depends on #50 because I didn't want to get into `String`-hell again :'-)

Will rebase once I get #50 finished.

### DIFF
--- a/libsignal-service/Cargo.toml
+++ b/libsignal-service/Cargo.toml
@@ -28,6 +28,7 @@ chrono = { version = "0.4", features = ["serde"] }
 log = "0.4"
 bincode = "1.3"
 
+sha-1 = "0.10"
 sha2 = "0.10"
 hmac = "0.12"
 hex = "0.4"
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -16,7 +16,8 @@ use crate::{
     proto::{ProvisionEnvelope, ProvisionMessage, ProvisioningVersion},
     provisioning::{ProvisioningCipher, ProvisioningError},
     push_service::{
-        AccountAttributes, HttpAuthOverride, PushService, ServiceError,
+        AccountAttributes, ContactTokenDetails, ContactTokenList,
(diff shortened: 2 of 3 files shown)


Be-ing:
#50 has been merged and this has some merge conflicts now. Could you rebase?

rubdos:
> #50 has been merged and this has some merge conflicts now. Could you rebase?

It's quite a bit of conflicts, sadly. I'll rebase when I work again on contact discovery; except if you want to continue this really soon?

Be-ing:
> if you want to continue this really soon?

Rebasing someone else's branch isn't a very nice way to get started in a new codebase.

rubdos:
So, I've rebased this now, but `retrieveDirectory` / `/v1/directory/tokens` seems like it's gone from the Signal client. We probably want to drop this and implement the new contact discovery protocol with attestation.

rubdos:
> the new contact discovery protocol with attestation.

Result of some investigative work:

Android's `ContactDiscoveryRefreshV1.java` contains the currently working logic to do contact lookup. However, Signal has been working on a V2, `ContactDiscoveryRefreshV2.kt` (notice Kotlin), tied to https://github.com/signalapp/ContactDiscoveryService. This version 2 has no endpoint on the production servers yet, but it has a lot of recent activity, and might replace V1 sooner than we want. Version 2 also has [some components in Signal-Client](https://github.com/signalapp/libsignal/tree/main/rust/attest), such that we don't have to implement everything ourselves.

So either we implement both, or we wait a bit longer until v2 is moved into production.

rubdos:
In other news: these Rust components in libsignal-client require `ecdsa` `^0.13`, which transitively depends on Rust 2021 and is as such currenly unusable in Whisperfish (because we're still on 1.52-dev, which on itself is not *that* old... come on library devs, please don't do this...)

Downgrading those versions leads to other impossibilities (`ecdsa`'s feature `serde` only exists in 0.13).

rubdos:
This stuff is completely deprecated and disfunctional by now. We should implement the new PNI system instead.

PR #57, opened 2021-01-14
Add some magic field when registering pre-keys

gferon:
This is necessary so the client is registered as compatible with groups v2.

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -109,11 +109,10 @@ impl<Service: PushService> AccountManager<Service> {
             pre_keys: pre_key_entities,
             signed_pre_key: signed_pre_key.into(),
             identity_key: identity_key_pair.public(),
-            // From Signal-Desktop: this is just to make the server happy (v2 clients should choke on publicKey)
-            last_resort_key: PreKeyEntity {
+            last_resort_key: Some(PreKeyEntity {
                 key_id: 0x7fffffff,
                 public_key: "NDI=".into(),
-            },
+            }),
         };
 
         self.service.register_pre_keys(pre_key_state).await?;
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -109,6 +109,11 @@ impl<Service: PushService> AccountManager<Service> {
             pre_keys: pre_key_entities,
             signed_pre_key: signed_pre_key.into(),
             identity_key: identity_key_pair.public(),
+            // From Signal-Desktop: this is just to make the server happy (v2 clients should choke on publicKey)
+            last_resort_key: PreKeyEntity {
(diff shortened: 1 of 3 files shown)


libsignal-service/src/pre_keys.rs
```diff
@@ -64,4 +64,5 @@ pub struct PreKeyState {
     pub signed_pre_key: SignedPreKey,
     #[serde(with = "serde_public_key")]
     pub identity_key: PublicKey,
+    pub last_resort_key: PreKeyEntity,
```
rubdos, line 67: If we merge this as-is, Whisperfish might break... Although I do intend to implement GV2 asap now, I'd like some compatibility flag, if possible.

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")] 
    pub last_resort_key: Option<PreKeyEntity>,
```

and a flag `last_resort_key: bool` in the method, maybe? This hack also does not seem to be present in `Signal-Android`, (cfr. `PushServiceSocket::registerPreKeys`, `PreKeyState`) would you know what that is about?

gferon: So, this is only for secondary devices, to not "block" the account to join groups v2 (at least, from my observations). If you look at `signal-cli` it's also done there: https://gitlab.stusta.de/AsamK/textsecure-cli/blob/067765a62f1877fced7a82fd86a800673e47df41/src/main/java/org/asamk/signal/Manager.java#L368

rubdos: If it's certainly only for secondary devices, I'd think this justifies the `Option`al and `serde` annotation, or am I getting that wrong?

gferon: You're absolutely right, I pushed a new commit, I think this should be good to merge now!

libsignal-service/src/account_manager.rs
```diff
@@ -109,6 +109,10 @@ impl<Service: PushService> AccountManager<Service> {
             pre_keys: pre_key_entities,
             signed_pre_key: signed_pre_key.into(),
             identity_key: identity_key_pair.public(),
+            last_resort_key: Some(PreKeyEntity {
+                key_id: 0x7fffffff,
+                public_key: "NDI=".into(),
+            }),
```
rubdos, line 115: This method still needs a way to signify that we want this `Some` or `None`; primary registrations needs to be able to put this at `None` (I'm using this method in WF).

rubdos (changes requested):

rubdos (changes requested):
`update_pre_key_bundle` method still needs a way to set `last_resort_key` to `None`

gferon:
> `update_pre_key_bundle` method still needs a way to set `last_resort_key` to `None`

Sorry, I didn't pay enough attention to your previous comments, I've added the flag now.

PR #58, opened 2021-01-19
Allow arbitrary Sinks to be passed to provision_secondary_device

ntninja:
Without this one cannot use combinators like `.with(|| …)` on created channels before passing them to this function.


rubdos:
This looks quite trivial, if you give rustfmt a listen (look at CI), I'll hit merge :-)

ntninja:
Fixed!

PR #59, opened 2021-01-23
Link devices: some extra methods

rubdos:
WIP for extra methods needed for device linking in Whisperfish https://gitlab.com/rubdos/whisperfish/-/merge_requests/61

Should be quite a trivial patch.


rubdos:
@gferon what do you think about 9abe73c? Introduces a dependency, but it's the de-facto standard, and I'd think most downstream users would use it anyway.

gferon (approved):

rubdos:
Confirmed working, merging!

PR #64, opened 2021-01-31
Additional sync message methods

rubdos:
- [x] Send contact sync message
  - [x] Upload contacts
- [x] Group sync message


rubdos:
CI failure because of `actix-rt v2.0.0-beta.3` being incompatible with `actix-http v3.0.0-beta.1`. Probably fixed in a few hours.

rubdos:
image

Tested contact sync with Whispelfish, that already works.

rubdos:
This is tested, and trivial enough not to warrant a real review, so I'm pulling this in :-)

PR #66, opened 2021-02-02
Implement captcha support when requesting verification codes

gamag:
Adds captcha and challenge support using two new methods in AccountManager  `request_sms_verification_code_with_data` and  `request_voice_verification_code_with_data` to keep the old interface working.

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -49,15 +49,40 @@ impl<Service: PushService> AccountManager<Service> {
         }
     }
 
+    pub async fn request_sms_verification_code_with_data(
+        &mut self,
+        phone_number: &str,
+        captcha: Option<&str>,
+        challenge: Option<&str>,
+    ) -> Result<SmsVerificationCodeResponse, ServiceError> {
+        Ok(self
+            .service
+            .request_sms_verification_code(phone_number, captcha, challenge)
+            .await?)
+    }
+
     pub async fn request_sms_verification_code(
         &mut self,
         phone_number: &str,
     ) -> Result<SmsVerificationCodeResponse, ServiceError> {
         Ok(self
             .service
-            .request_sms_verification_code(phone_number)
+            .request_sms_verification_code(phone_number, None, None)
             .await?)
     }
+
+    pub async fn request_voice_verification_code_with_data(
+        &mut self,
+        phone_number: &str,
+        captcha: Option<&str>,
+        challenge: Option<&str>,
+    ) -> Result<VoiceVerificationCodeResponse, ServiceError> {
+        Ok(self
+            .service
+            .request_voice_verification_code(phone_number, captcha, challenge)
+            .await?)
+    }
+
 
     pub async fn request_voice_verification_code(
         &mut self,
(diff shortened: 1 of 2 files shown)


libsignal-service/src/push_service.rs
```diff
@@ -350,32 +350,85 @@ pub trait PushService {
         file: Option<(&str, &'s mut C)>,
     ) -> Result<(), ServiceError>;
 
+    fn build_verification_code_request_url(
+        msg_type: &str,
+        phone_number: &str,
+        captcha: Option<&str>,
+        challenge: Option<&str>,
```
rubdos, line 357: Doesn't it make more sense to have this in two/three separate methods, since only one of them can be `Some`, and none of these is actually dynamically dispatched?

libsignal-service/src/account_manager.rs
```diff
@@ -49,13 +49,37 @@ impl<Service: PushService> AccountManager<Service> {
         }
     }
 
+    pub async fn request_sms_verification_code_with_data(
```
rubdos, line 52: Feel free to break interfaces. @gferon and us are the main consumers of the library, and we break the interface all the time.

rubdos (commented):
Some minor things/discussion points.

libsignal-service/src/push_service.rs
gamag: For the implementation here that would mean quite some code duplication, I think. We could get rid of the url formatting method though - I keep forgetting that we can assign complex if statements. For the account manager that could make more sense maybe.

libsignal-service/src/account_manager.rs
gamag: Good to know. Do you care about keeping this close to the java lib interface? If not, I'll split the method in two again.

rubdos: I like to keep the reference (as in: comments to where to find the respective Java methods), but if I have to pick between something Rusty and something 1:1 compatible, I'm 100% in camp Rust.

gamag: so would be different methods more rusty?

shekohex:
Sorry, but this could be off-topic here but does anyone have a reference for where Signal Server implement this feature? sending a captcha challenge when trying to register?

because I'm looking to implement something similar.

libsignal-service/src/account_manager.rs
rubdos: Both method names would give away the intent of the call: there can be a `request_captcha` method, and then you can alter the original method (`request_sms`) with an argument `captcha: Option<String>`, or would that not make sense?

rubdos:
> Sorry, but this could be off-topic here but does anyone have a reference for where Signal Server implement this feature? sending a captcha challenge when trying to register?

https://github.com/signalapp/Signal-Server/blob/master/service/src/main/java/org/whispersystems/textsecuregcm/controllers/AccountController.java ^F captcha, afaict. It's using recaptcha.

libsignal-service/src/account_manager.rs
gamag: Not sure I understand what you mean: I think there should be either three methods per type (sms/voice) `request_sms_verification_code_with_captcha`, `request_sms_verification_code_with_challenge` and `request_sms_verification_code` or only the last one with two optional arguments ...

gamag: removed the "with data" methods. Seemed to be the easiest option for whisperfish to use.

rubdos:
LGTM, let's get this all over with :-)

PR #69, opened 2021-02-25
Add device ID to credentials and use it in login

### LINKED ISSUE #68: Unable to get authorized with a non-default device ID after linking
First of all, nice project! I've been poking around to use it for my own Signal CLI utility.

However, I have been unable to get authorized with the push service after linking a device and instantiating a new push service with appropriately set service config and credentials (i.e. all requests end up with `401 Unauthorized`).

After some more investigation (and after looking at the forked [libsignal-service-java](https://github.com/Turasa/libsignal-service-java/blob/bdd1c753d67a2552d24865dee94c55e5148d55b2/service/src/main/java/org/whispersystems/signalservice/internal/push/PushServiceSocket.java#L1871)), I came to understand that the authorization header should also include the device ID, in case the request is being made on behalf of non-primary device. I.e. for non-primary device, the header should be composed in format `{identifier}.{device_id}:{password}`.

Locally, I was able to make the authorization work by building the login with the said device ID. 

I was wondering whether this is the right solution to my problem (or whether I'm doing something incorrectly from the beginning). If so, I'd be also happy to provide a PR for this.

matej-g:
As was described in #68, currently the users of the service are not able to authorize themselves when communicating with Signal servers, when making requested on behalf of a non-primary device (e.g. after linking another device via the service). The expected authorization header should in such case be built in format `{identifier}.{device_id}:{password}`.
 
In particular, this PR:
- Adds `device_id: Option<i32>` to `Credentials`
- If `device_id` is set to non-primary device ID (i.e. other than `1`), build the `login` for authorization header in format `{identifier}.{device_id}`

Resolves #68

### DIFF
--- a/libsignal-service/src/configuration.rs
+++ b/libsignal-service/src/configuration.rs
@@ -4,7 +4,7 @@ use libsignal_protocol::{keys::PublicKey, Context};
 
 use crate::{
     envelope::{CIPHER_KEY_SIZE, MAC_KEY_SIZE},
-    push_service::{DEFAULT_DEVICE_ID, ServiceError},
+    push_service::{ServiceError, DEFAULT_DEVICE_ID},
     sealed_session_cipher::{CertificateValidator, SealedSessionError},
 };
 
@@ -43,14 +43,13 @@ impl Credentials {
                 uuid
             } else {
                 &self.e164
-            }.to_owned()
+            }
+            .to_owned()
         };
 
         match self.device_id {
             None | Some(DEFAULT_DEVICE_ID) => identifier,
-            Some(id) => {
-                identifier + "." + id.to_string().as_str()
-            },
+            Some(id) => identifier + "." + id.to_string().as_str(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


libsignal-service/src/configuration.rs
```diff
@@ -25,22 +25,31 @@ pub struct Credentials {
...
+                uuid
+            } else {
+                &self.e164
+            }
+            .to_owned()
+        };
+
+        match self.device_id {
+            None | Some(DEFAULT_DEVICE_ID) => identifier,
+            Some(id) => identifier + "." + id.to_string().as_str(),
```
rubdos, line 52:
```suggestion
            Some(id) => format!("{}.{}", identifier, id),
```

Looks cleaner I think.

rubdos (commented):
Small remark. Otherwise looks good to me. I'd still like to know what @gferon does for this though.

libsignal-service/src/configuration.rs
gferon: > I'd still like to know what @gferon does for this though.

Well, you got it exactly right - I do it exactly like that, except I never somehow thought it could be integrated directly in the lib (see: https://github.com/gferon/presage/blob/wip/src/manager.rs#L112)

Thanks a lot for this improvement!

gferon (approved):

libsignal-service/src/configuration.rs
matej-g: Glad I could contribute!

Issue #47, opened 2020-11-11
Use strong types for Uuid and E164


gferon:
Right now, we both handle phone numbers (e164) and uuid as strings.

Replace this by using: [`Uuid`](https://crates.io/crates/uuid) and a `PhoneNumber` (containing a `String`) that can validate input with a crate like [`phonenumber`](https://crates.io/crates/phonenumber).

PR #71, opened 2021-02-26
Re-export some useful types

gferon:
Useful to not force users to depend on the crate

Relates to #47

### DIFF
--- a/libsignal-service/src/lib.rs
+++ b/libsignal-service/src/lib.rs
@@ -43,5 +43,5 @@ pub mod prelude {
         receiver::MessageReceiver,
         sender::{MessageSender, MessageSenderError},
     };
-    pub use phonenumber::{parse, PhoneNumber};
+    pub use phonenumber::{parse as parse_phonenumber, PhoneNumber};
 }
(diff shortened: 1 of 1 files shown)


libsignal-service/src/lib.rs
```diff
@@ -43,4 +43,5 @@ pub mod prelude {
         receiver::MessageReceiver,
         sender::{MessageSender, MessageSenderError},
     };
+    pub use phonenumber::{parse, PhoneNumber};
```
rubdos, line 46:
```suggestion
    pub use phonenumber::{parse as parse_phonenumber, PhoneNumber};
```

Maybe? Because otherwise we're having a loose `parse()` function in our `prelude`.

rubdos:
Could also try just `pub use phonenumber;` in prelude, but that's maybe less clear.

gferon:
> Could also try just `pub use phonenumber;` in prelude, but that's maybe less clear.

Yeah, you're right. I also realized we would also need to re-export `uuid`.

shekohex (approved):

rubdos:
Were you going to re-export the uuid in here already too?

rubdos:
@gferon, this happened some place else I suppose, should be drop the branch?

gferon:
> @gferon, this happened some place else I suppose, should be drop the branch?

Yes!

Issue #63, opened 2021-01-28
Code duplication in Service implementations.


rubdos:
get/put/post/delete share a whole lot of code, but they differ ever so slightly in their implementation. Should probably be rewired through a macro.

gferon:
It should already be improved by work in #73 and the future `hyper` implementation has even more shared code compared to the `actix` one #spoiler

PR #73, opened 2021-03-02
Initial work towards supporting Groups v2

gferon:
I open this PR so we can start discussing about the general direction we want to take for this change.

Expect: dirty code, hacks, ugly things and... being able to fetch the metadata of a group v2 given that you posses its `master_key` (which is part of what the client receives on _every_ new message).

I'd mostly like to get some early review, and answers to the following questions:

* Is the trait for `CredentialsCache` ok? (It's modeled after the cache used in the Java source code)
* `awc::Client` sets  the `Authorization` header twice if it's set at the client and request level, so I had to remove it from `get_client` and set it manually everywhere. This is error prone, but I couldn't think of another way
* Are there too many `.unwrap()` in this change? 💝 

TODO:
- [x] Decrypt the incoming data
- [x] Clean things up


libsignal-service/src/gv2/api.rs
```diff
@@ -0,0 +1,156 @@
...
+pub trait CredentialsCache {
+    fn clear(&self) -> Result<(), CredentialsCacheError>;
+    fn read(
+        &self,
+    ) -> Result<HashMap<i64, AuthCredentialResponse>, CredentialsCacheError>;
+    fn write(
+        &self,
+        value: &HashMap<i64, AuthCredentialResponse>,
+    ) -> Result<(), CredentialsCacheError>;
+}
```
rubdos, line 36: I think I'd rather see a `read(i16) -> Result<ACR>; write(i64, ACR)` for this, possibly with a read_many and write_many if really necessary.

rubdos: The read/write_many can have a default-impl, and should probably take some iterator forms...

rubdos: We can also supply an in-memory and in-xdg-cache implementation for these!

libsignal-service/src/gv2/api.rs
```diff
@@ -0,0 +1,156 @@
...
+        today.num_days()
+    }
+
+    fn get_authorization_string(
+        &self,
+        uuid: Uuid,
+        group_secret_params: GroupSecretParams,
+        credential_response: AuthCredentialResponse,
+        today: u32,
+    ) -> Result<(String, String), ServiceError> {
```
rubdos, line 115: Maybe consider a strong type for user/password?

libsignal-service/src/gv2/api.rs
```diff
@@ -0,0 +1,156 @@
...
+        let auth_credential = self
+            .server_public_params
+            .receive_auth_credential(
+                *uuid.as_bytes(),
+                today,
+                &credential_response,
+            )
+            .unwrap();
```
rubdos, line 123: This sounds fallible, no? Otherwise, maybe document why this unwrap is safe, possibly by using .expect instead.

libsignal-service/src/gv2/models.rs
```diff
@@ -0,0 +1,30 @@
+use std::collections::HashMap;
+
+use serde::Deserialize;
+use zkgroup::auth::AuthCredentialResponse;
+
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TemporalCredential {
+    credential: String,
+    redemption_time: i64,
```
rubdos, line 10: Should we use `chrono::DateTime` to signify when it expires instead?

libsignal-service-actix/src/push_service.rs
```diff
@@ -26,30 +27,36 @@ impl PushService for AwcPushService {
...
+                ServiceError::Timeout {
                     reason: e.to_string(),
-                },
+                }
+            }
+            _ => ServiceError::SendError {
+                reason: e.to_string(),
             },
-        )?;
+        })?;
```
rubdos, line 59: This code duplication between all the get/post/put methods should probably be macro'd away. That's probably for #63. That should also get rid of your concern for adding the authorization header manually...

libsignal-service/src/configuration.rs
```diff
@@ -104,41 +116,39 @@ impl Into<ServiceConfiguration> for SignalServers {
...
                 },
-                contact_discovery_url: vec![
-                    "https://api-staging.directory.signal.org".into(),
-                ],
+                contact_discovery_url:
+                    "https://api-staging.directory.signal.org".parse().unwrap(),
                 certificate_authority: SIGNAL_ROOT_CA.into(),
                 unidentified_sender_trust_root:
                     "BbqY1DzohE4NUZoVF+L18oUPrK3kILllLEJh2UnPSsEx".into(),
+                zkgroup_server_public_params: bincode::deserialize(&base64::decode("ABSY21VckQcbSXVNCGRYJcfWHiAMZmpTtTELcDmxgdFbtp/bWsSxZdMKzfCp8rvIs8ocCU3B37fT3r4Mi5qAemeGeR2X+/YmOGR5ofui7tD5mDQfstAI9i+4WpMtIe8KC3wU5w3Inq3uNWVmoGtpKndsNfwJrCg0Hd9zmObhypUnSkfYn2ooMOOnBpfdanRtrvetZUayDMSC5iSRcXKpdls=").unwrap()).unwrap(),
```
rubdos, line 132: Is there a nicer way to represent the public params? Probably not immediately...

rubdos (commented):
Just some things that caught my eye.

libsignal-service/src/gv2/api.rs
gferon: so... it doesn't really help to expose the internal hash map because it works like this:
1. Get 7 days of credentials
2. Use the cache
3. At some point, you don't have a new day in the cache, trash everything, and write the next 7 days

Since the current back-end I have is `sled`, it does make sense to have key-value in the interface, but it's not really useful as far as the actual feature is concerned.

libsignal-service/src/gv2/models.rs
gferon: This is what feeds the credential in the `HashMap<i64, ...>` and directly comes from the API, so I'm not sure there's any benefit to it?

libsignal-service-actix/src/push_service.rs
gferon: I ended up using `client::request` which feels much better now!

libsignal-service/src/configuration.rs
gferon: Unfortunately not... [`ServerPublicParams` doesn't expose any way to instantiate it without using `serde::Deserialize`](https://github.com/signalapp/zkgroup/blob/master/rust/src/api/server_params.rs#L19)

libsignal-service/src/gv2/api.rs
rubdos: Is this API per-group, or per-client? It's just weird to have a `read()` without parameters (except for self)

gferon: It's per-client, so let's say "global".

rubdos: And the cached data is for all groups together then? I'd think there's a cached authorization proof per group.

gferon: You're absolutely right, I got mixed up in the onion layers.

libsignal-service/src/gv2/models.rs
```diff
@@ -0,0 +1,30 @@
...
+    credentials: Vec<TemporalCredential>,
+}
+
+impl CredentialResponse {
+    pub fn parse(self) -> HashMap<i64, AuthCredentialResponse> {
+        self.credentials
+            .into_iter()
+            .map(|c| {
+                let bytes = base64::decode(c.credential).unwrap();
+                let data = bincode::deserialize(&bytes).unwrap();
```
rubdos, line 25: Fallible x2

libsignal-service/src/gv2/operations.rs
```diff
@@ -0,0 +1,221 @@
...
+        })
+    }
+
+    fn decrypt_pending_member(
+        &self,
+        member: EncryptedPendingMember,
+    ) -> Result<DecryptedPendingMember, GroupDecryptionError> {
+        let inner_member =
+            member.member.ok_or(GroupDecryptionError::WrongBlob)?;
+        let uuid = self.decrypt_uuid(&inner_member.user_id).unwrap_or_default();
```
rubdos, line 96: Maybe document why having `[0u8; _]` here is better than throwing an error for the decryption.

rubdos (commented):
Doesn't look to bad tbh.

Maybe think about the `api.rs` naming, because everything is API in the whole crate.

libsignal-service/src/gv2/api.rs
```diff
@@ -0,0 +1,195 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum CredentialsCacheError {
+    #[error("failed to read values from cache: {0}")]
+    ReadError(String),
+    #[error("failed to write values from cache: {0}")]
+    WriteError(String),
+}
+
+pub trait CredentialsCache {
```
rubdos, line 29: Can do with some documentation now, maybe some example code in the doc string.

libsignal-service/src/gv2/api.rs
```diff
@@ -0,0 +1,195 @@
...
+        let mut random_bytes = [0u8; 32];
+        rand::thread_rng().fill_bytes(&mut random_bytes);
+
+        let auth_credential_presentation = self
+            .server_public_params
+            .create_auth_credential_presentation(
+                random_bytes,
+                group_secret_params,
+                auth_credential,
+            );
```
rubdos, line 167: This is something we may want to hand over to a threadpool, but that's of course backend-dependent...

rubdos (commented):
*empty review*, that's not useful.

libsignal-service/src/gv2/operations.rs
```diff
@@ -0,0 +1,221 @@
...
+                        .map_err(GroupDecryptionError::ProtobufDecodeError)
+                })
+                .unwrap_or_else(|e| {
+                    log::warn!("bad encrypted blob: {}", e);
+                    GroupAttributeBlob::default()
+                })
+        }
+    }
+
+    fn decrypt_title(
```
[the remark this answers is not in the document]
gferon: I'd really like to avoid macros as much as we can, as it makes browsing the code just... painful. What do you mean exactly by scaffold? One thing we might want to do is directly `impl` things on the protobuf mirrored structs themselves (since it's included in the crate, we should be able to.)

libsignal-service/src/gv2/operations.rs
```diff
@@ -0,0 +1,221 @@
...
+        } else {
+            use group_attribute_blob::Content;
+            self.group_secret_params
+                .decrypt_blob(bytes)
+                .map_err(|_| GroupDecryptionError::ZkGroupError)
+                .and_then(|b| {
+                    GroupAttributeBlob::decode(Bytes::copy_from_slice(&b[4..]))
+                        .map_err(GroupDecryptionError::ProtobufDecodeError)
+                })
+                .unwrap_or_else(|e| {
```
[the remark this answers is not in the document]
gferon: This is the behavior of the original implementation, it looks like they ignore those errors on purpose, so I was thinking we should keep it like this?

libsignal-service/src/gv2/api.rs
gferon: At this point, since `actix` is based on `tokio` I'd just use `tokio::spawn_blocking` and `.await` on the handle. But in general, we do operations like this everywhere, don't we? (so this doesn't really apply to only here)

gferon:
> Doesn't look to bad tbh.
> 
> Maybe think about the `api.rs` naming, because everything is API in the whole crate.

What about having everything in `mod.rs`? It's really not that much code.

rubdos:
> > Doesn't look to bad tbh.
> > Maybe think about the `api.rs` naming, because everything is API in the whole crate.
> 
> What about having everything in `mod.rs`? It's really not that much code.

That, or modeled after `AccountManager` we could have `group_manager.rs`?

libsignal-service/src/gv2/operations.rs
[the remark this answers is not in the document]
rubdos: It's just all the same code with different function names, that's why I suggested a macro. It can probably be done with enum and visitor too, but don't feel pressured into refactoring this yet. We can do it later.

libsignal-service/src/gv2/api.rs
rubdos: Let's postpone until we decide to merge actix and hyper as features into the main crate. That'll make more sense at that point. Actix has its own threadpool, FWIW.

libsignal-service/src/gv2/operations.rs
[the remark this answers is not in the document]
gferon: Hmm ok, I'd say that if the code duplicates more than x2 (it's currently exactly 2) we should look into a macro. I think it'll only show when we implement other protobuf structs and I'd do it then.

libsignal-service/src/gv2/api.rs
gferon: since now we have `InMemoryCredentialsCache` it shouldn't be a problem, I added a comment though.

libsignal-service/src/groups_v2/manager.rs
```diff
@@ -0,0 +1,231 @@
...
+        key: &i64,
+    ) -> Result<Option<&AuthCredentialResponse>, CredentialsCacheError> {
+        Ok(self.map.get(key))
+    }
+
+    fn write(
+        &mut self,
+        mut map: HashMap<i64, AuthCredentialResponse>,
+    ) -> Result<(), CredentialsCacheError> {
+        std::mem::swap(&mut self.map, &mut map);
```
rubdos, line 101: Fancy pants. What's wrong with self.map = map ? :'-)
```suggestion
        self.map = map;
```

(this will make it unnesessary for map to be mut too)

libsignal-service-actix/src/push_service.rs
```diff
@@ -210,24 +293,76 @@ impl PushService for AwcPushService {
...
+        endpoint: Endpoint,
+        path: &str,
+        value: S,
+    ) -> Result<D, ServiceError>
+    where
+        D: Default + ProtobufMessage,
+        S: Sized + ProtobufMessage,
+    {
+        let mut buf = vec![];
+        value.encode(&mut buf)?;
```
rubdos, line 334: `with_capacity` and `.expect()` until we have https://github.com/danburkert/prost/pull/378

libsignal-service/src/push_service.rs
```diff
@@ -271,10 +278,25 @@ pub enum ServiceError {
...
+
+    #[error("Error decoding response: {reason}")]
+    ResponseError { reason: String },
+
     #[error("Error decoding JSON response: {reason}")]
     JsonDecodeError { reason: String },
+    #[error("Error decoding protobuf frame: {0}")]
+    ProtobufDecodeError(#[from] prost::DecodeError),
+    #[error("Error encoding protobuf frame: {0}")]
+    ProtobufEncodeError(#[from] prost::EncodeError),
```
rubdos, line 295: This shouldn't be necessary if you replace all the encode(&mut Vec) stuff with `with_capacity` and `.expect()`. Encoding is infallible except when the vector doesn't grow. See https://github.com/danburkert/prost/pull/378 :-)

rubdos (approved):
Some last very minor nitpicks. Thanks for the work!

PR #74, opened 2021-03-08
Add libsignal-service implementation with Hyper, Tungstenite and Tokio

gferon:


libsignal-service/Cargo.toml
```diff
@@ -43,3 +43,5 @@ anyhow = "1.0"
 
 [features]
 prefer-e164 = []
+actix = []
+hyper = []
```
rubdos, line 47: Did we already want to handle this here?

rubdos (commented):
I'm not gonna read the whole Hyper thingy, it looks alright ;p

Other than that, we probably want the hyper/actix features in a separate MR?

libsignal-service/Cargo.toml
gferon: You're right, this is a remnant from... the future. I'll probably open a MR to use features right after this one is merged, but only if it makes sense after I'm done experimenting.

gferon:
@rubdos if it's fine with you, I'll squash and merge this one.

PR #81, opened 2021-04-19
De-duplicate provisioning code

gferon:
These are changes that I believe are necessary to move `provision_secondary_device` away from the concrete implementations of the library (`actix`, `hyper`, etc.).

Since credentials are always handled weirdly, I started to experiment with an idea that hides `Credentials` creation for API users, and expose only the fields that are relevant when instantiating a `ProvisioningManager` or `LinkingManager` - thus avoiding any potential mistakes.

I opened this PR mostly to discuss the approach, and merge it after iterating if those changes are desirable.

Changes of this PR in a nutshell:
- Add a `fn new(cfg, credentials, user_agent)` method to the `PushService` trait so that abstract code can create a new `PushService` with a different set of credentials, see `ProvisioningManager::new` and `LinkingManager::new`.
- Add the ability to override credentials in `get_json` and `put_json`


libsignal-service-actix/src/push_service.rs
```diff
@@ -120,6 +106,20 @@ impl PushService for AwcPushService {
...
+    ) -> Self {
+        let cfg = cfg.into();
+        let client = get_client(&cfg, user_agent);
+        Self {
+            cfg,
+            credentials: credentials.and_then(|c| c.authorization()),
+            client,
+        }
+    }
+
```
rubdos, line 122: For the purposes of a clean diff, you may want to move this up a bit.

libsignal-service-actix/src/push_service.rs
```diff
@@ -120,6 +106,20 @@ impl PushService for AwcPushService {
     type ByteStream = Box<dyn futures::io::AsyncRead + Unpin>;
     type WebSocket = AwcWebSocket;
 
+    fn new(
+        cfg: impl Into<ServiceConfiguration>,
+        credentials: Option<ServiceCredentials>,
+        user_agent: &'static str,
```
rubdos, line 112: Can we make this less stringent than `&'static str`? Whisperfish will almost certainly provide a way to dynamically set a UA.

libsignal-service/src/provisioning/cipher.rs
```diff
@@ -0,0 +1,223 @@
...
+use crate::{
+    envelope::{CIPHER_KEY_SIZE, IV_LENGTH, IV_OFFSET},
+    provisioning::ProvisioningError,
+};
+
+#[derive(Debug)]
+enum CipherMode {
+    Decrypt(KeyPair),
+    Encrypt(PublicKey),
+}
```
rubdos, line 27: Still a bit unclear tbh., but changing this to `DecryptAndEncrypt` and `EncryptOnly` would probably be better. Not part of this PR though, I suppose. (BTW, Rust Analyzer has a great rename function for this).

rubdos (changes requested):
From afar, looking pretty good. Two minor remark wrt. lifetimes and wrt to diff cleanliness.

rubdos:
> Since credentials are always handled weirdly, I started to experiment with an idea that hides `Credentials` creation for API users, and expose only the fields that are relevant when instantiating a `ProvisioningManager` or `LinkingManager` - thus avoiding any potential mistakes.

This probably also needs doing in `AccountManager`, but I've always felt dirty too with those things. `AccountManager` has the additional problem that setting `profile_key` is needed for some methods but not for others. We might even want to make that a compile-time check with a generic...

libsignal-service-actix/src/push_service.rs
gferon: Hmm, not sure what the best to achieve this is, since it's moved from `impl AwcPushService` to `impl PushService for AwcPushService`.

gferon:
> We might even want to make that a compile-time check with a generic...

I agree, should we open an issue to do it in another PR?

libsignal-service-actix/src/push_service.rs
rubdos: Ack, not a problem.

rubdos:
> I agree, should we open an issue to do it in another PR?

Yes, definitely. Not soon, because I'm touching that code again in the libsignal-client rewrite.

rubdos (approved):
Mergable, I think.

PR #82, opened 2021-04-19
Upgrade actix crates (with fixes)

gferon:
Should fix compilation errors when you don't have the right `Cargo.lock`... for now!

`awc` > `3.0.0-beta.5`
`actix-http` > `3.0.0-beta.6`


rubdos:
Not a lot to say atm; if you're not pushing more than this, go ahead and merge!

PR #83, opened 2021-04-20
Swap libsignal-protocol dependency with libsignal-client

### LINKED ISSUE #72: Consider/switch to libsignal-client
instead of `libsignal-protocol-c`. It's already used in production in (at least) the Android app.

Link: https://github.com/signalapp/libsignal-client

rubdos:
Fixes #72


rubdos:
It's a bit of a rewrite, so I'm pushing broken commits as I pass over code. In the end I'll squash the `part x` things to something with a real commit message :-)

rubdos:
Important detail. `libsignal-client` is AGPLv3. We have to move to AGPLv3 from this point.

Be-ing:
No, relicensing to AGPL is not needed and can't be done without explicit agreement of everyone who has contributed to this repo:

> The ordinary GNU General Public License and the GNU Affero General Public License are two different copyleft licenses, so they are naturally incompatible. We have set up a special kind of explicit compatibility between them: you can include source code under the GNU GPL version 3 together with other source code under the GNU Affero GPL in a single combined program. This is permitted because both of those licenses explicitly say so, and the effect is that the GNU AGPL applies to the combined program. However, you can't simply relicense code from the GNU GPL (with or without “or later”) to the GNU Affero GPL, or vice versa; neither of these licenses gives permission for that. Note also that the GNU Affero GPL version 3 is not a “later version” of the ordinary GNU GPL version 2, because the GNU Affero GPL and the GNU GPL are two different series of licenses.

https://www.gnu.org/licenses/license-compatibility.html

rubdos:
So either we relicense the whole thing, or we go the complicated route as I took with Whisperfish: old source is GPLv3 and new source is AGPLv3. If we don't get the permission of them all, that's what we should do.

I'll make an announcement on issue #72.

Either way, the whole of libsignal-service-rs will fall under AGPLv3, since it's a combination of (A)GPLv3 and AGPLv3 sources.

libsignal-service/Cargo.toml
```diff
@@ -7,8 +7,8 @@ license = "GPLv3"
 readme = "../README.md"
 
 [dependencies]
-libsignal-protocol = { git = "https://github.com/Michael-F-Bryan/libsignal-protocol-rs" }
-zkgroup = { git = "https://github.com/signalapp/zkgroup" }
+libsignal-protocol = { git = "https://github.com/whisperfish/libsignal-client", branch = "make-error-sync" }
```
rubdos, line 10: Should we wait on https://github.com/signalapp/libsignal-client/pull/279?

libsignal-service/src/account_manager.rs
```diff
@@ -78,11 +75,14 @@ impl<Service: PushService> AccountManager<Service> {
     /// Equivalent to Java's RefreshPreKeysJob
     ///
     /// Returns the next pre-key offset and next signed pre-key offset as a tuple.
-    pub async fn update_pre_key_bundle(
+    pub async fn update_pre_key_bundle<R: rand::Rng + rand::CryptoRng>(
         &mut self,
-        store_context: StoreContext,
+        identity_store: &dyn IdentityKeyStore,
+        pre_key_store: &mut dyn PreKeyStore,
+        signed_pre_key_store: &mut dyn SignedPreKeyStore,
```
rubdos, line 82: Let's make an issue about Rustyfing these. Or do we want a `-service`-exposed `StoreContext` that encompasses those three?

FWIW, `StoreContext` basically was a `Box<dyn T>` implementation too :'-)

libsignal-service/src/account_manager.rs
```diff
@@ -78,11 +75,14 @@ impl<Service: PushService> AccountManager<Service> {
...
+        csprng: &mut R,
         pre_keys_offset_id: u32,
-        next_signed_pre_key_id: u32,
+        signed_pre_key_id: u32,
```
rubdos, line 85: Does this mean that we have to change our call to this method?

libsignal-service/src/account_manager.rs
```diff
@@ -99,34 +99,54 @@ impl<Service: PushService> AccountManager<Service> {
...
+            .private_key()
+            .calculate_signature(&signed_pre_key_public.serialize(), csprng)?;
+
+        let unix_time = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap();
+
+        let signed_prekey_record = SignedPreKeyRecord::new(
+            signed_pre_key_id + 1,
+            unix_time.as_secs(),
```
rubdos, line 133: Signal seems to like milliseconds for their timestamps, are we sure it's seconds here?

rubdos: Honestly, ::new should accept a DateTime instead :'-)

libsignal-service/src/cipher.rs
```diff
@@ -10,60 +22,58 @@ use crate::{
...
     ) -> Result<Option<Content>, ServiceError> {
         if envelope.legacy_message.is_some() {
-            let plaintext = self.decrypt(&envelope)?;
+            let plaintext = self.decrypt(&envelope).await?;
             let message =
                 crate::proto::DataMessage::decode(plaintext.data.as_slice())?;
             Ok(Some(Content::from_body(message, plaintext.metadata)))
         } else if envelope.content.is_some() {
-            let plaintext = self.decrypt(&envelope)?;
+            let plaintext = self.decrypt(&envelope).await?;
```
rubdos, line 76: FWIW, we need to discuss whether decrypting envelopes is still necessary. OWS stopped using the signaling key for encrypting envelopes, apparently.

libsignal-service/src/lib.rs
```diff
@@ -59,4 +60,13 @@ pub mod prelude {
...
     pub use zkgroup::groups::{GroupMasterKey, GroupSecretParams};
+
+    pub mod protocol {
+        pub use libsignal_protocol::{
+            Context, Direction, IdentityKey, IdentityKeyPair, IdentityKeyStore,
+            KeyPair, PreKeyRecord, PreKeyStore, PrivateKey, ProtocolAddress,
+            PublicKey, SessionRecord, SessionStore, SignalProtocolError,
+            SignedPreKeyRecord, SignedPreKeyStore,
+        };
+    }
```
rubdos, line 71: Good thinking!!

libsignal-service/src/push_service.rs
```diff
@@ -574,35 +596,15 @@ pub trait PushService {
...
-            bundle = bundle.signature(&signed_pre_key.signature);
-        }
-        if let Some(pre_key) = device.pre_key {
-            bundle = bundle.pre_key(
-                pre_key.key_id,
-                &PublicKey::decode_point(context, &pre_key.public_key)?,
-            );
-        }
-        Ok(bundle.build()?)
+        Ok(device.into_bundle(identity)?)
```
rubdos, line 601: @gferon, iirc I wrote this, could you check whether this seems alright wrt. `impl PreKeyResponseItem {  fn into_bundle(` a bit higher up?

libsignal-service/src/sender.rs
```diff
@@ -461,21 +477,11 @@ where
...
                         }
-                        if let Some(e164) = recipient.e164() {
-                            self.cipher.store_context.delete_session(
-                                &libsignal_protocol::Address::new(
-                                    &e164,
-                                    *extra_device_id,
-                                ),
-                            )?;
+                        if let Some(_e164) = recipient.e164() {
+                            unimplemented!("deleting session: unimplemented following the switch to the Rust version of libsignal-protocol");
```
rubdos, line 484: Tracking issue for this? This definitely affects Whisperfish.

libsignal-service/src/sender.rs
```diff
@@ -634,30 +638,27 @@ where
...
+                self.session_store
                     .get_sub_device_sessions(&uuid.to_string())?,
             );
         }
         if let Some(e164) = &recipient.e164() {
-            sub_device_sessions.extend(
-                self.cipher.store_context.get_sub_device_sessions(&e164)?,
-            );
+            sub_device_sessions
+                .extend(self.session_store.get_sub_device_sessions(&e164)?);
```
rubdos, line 650: `recipient.get_sub_device_sessions(session_store)` would be nice to have, which is what I meant with that comment.

rubdos (commented):
Looks like we're not far off! :-)

libsignal-service/Cargo.toml
rubdos: Yes we will wait.

libsignal-service/src/sender.rs
gferon: Weird, I swear I reverted that. I added `SessionStoreExt` so that's possible again.

libsignal-service/src/account_manager.rs
gferon: Yes, the problem is just what happens to the type when you want to require `Clone` and push the responsibility of handling concurrent mutability to the implementer. If you wrap those in a `StoreContext`, you can't easily derive `Clone` (at least, I tried, and... 😪)

Leptopoda:
> So either we relicense the whole thing, or we go the complicated route as I took with Whisperfish: old source is GPLv3 and new source is AGPLv3. If we don't get the permission of them all, that's what we should do.
> 
> I'll make an announcement on issue #72.
> 
> Either way, the whole of libsignal-service-rs will fall under AGPLv3, since it's a combination of (A)GPLv3 and AGPLv3 sources.
> 
> And wrt. sending to libsignal-client: that's a whole other story, because they require a CLA. AFAICT, for moving code, only my code and @gferon's code is being moved on the short term.

Just FYI that's exactly the reason why they have the CLA. They want to own the code so they can release it under a different license or sell it as closed source (like they did to Facebook or so)

rubdos:
> Just FYI that's exactly the reason why they have the CLA. They want to own the code so they can release it under a different license or sell it as closed source (like they did to Facebook or so)

Yes, and that's exactly the reason that I'm against CLA's too. In my opinion, the person that wrote the code should have the right to choose whether a change of license is okay with them, like in this case :-)

But let's stick to topic, we're very close having `libsignal-client` replacing `libsignal-protocol-rs`! :tada:

gferon:
Maybe you mean `libsignal-protocol` by `libsignal-protocol`? 😂 Both the pure C-bindings crate and the newer one have the same name.

rubdos:
potato tomato!

gferon:
@rubdos I think this is ready for a second (and hopefully final?) look!

libsignal-service/Cargo.toml
```diff
@@ -7,8 +7,8 @@ license = "GPLv3"
 readme = "../README.md"
 
 [dependencies]
-libsignal-protocol = { git = "https://github.com/Michael-F-Bryan/libsignal-protocol-rs" }
-zkgroup = { git = "https://github.com/signalapp/zkgroup" }
+libsignal-protocol = { git = "https://github.com/signalapp/libsignal-client" }
```
Be-ing, line 10:
```suggestion
libsignal-client = { git = "https://github.com/signalapp/libsignal-client" }
```
It is confusing that this is a different name than the repo.

rubdos: Not ours to choose. The repo contains several crates, among which `libsignal-protocol`.

libsignal-service/src/account_manager.rs
rubdos: Let's track this at #92

libsignal-service/src/account_manager.rs
```diff
@@ -78,13 +75,16 @@
     /// Equivalent to Java's RefreshPreKeysJob
     ///
     /// Returns the next pre-key offset and next signed pre-key offset as a tuple.
-    pub async fn update_pre_key_bundle(
+    pub async fn update_pre_key_bundle<R: rand::Rng + rand::CryptoRng>(
         &mut self,
-        store_context: StoreContext,
+        identity_store: &dyn IdentityKeyStore,
+        pre_key_store: &mut dyn PreKeyStore,
+        signed_pre_key_store: &mut dyn SignedPreKeyStore,
```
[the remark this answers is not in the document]
rubdos: #92

libsignal-service/src/cipher.rs
```diff
@@ -10,60 +22,65 @@ use crate::{
...
-                context,
-                store_context,
-                local_address,
-                certificate_validator,
+                session_store,
+                identity_key_store,
+                signed_pre_key_store,
+                pre_key_store,
+                csprng,
+                certificate_validator.clone(),
```
rubdos, line 65:
```suggestion
                certificate_validator,
```

libsignal-service/src/push_service.rs
rubdos: @gferon, can you verify whether this still checks out?

libsignal-service/src/push_service.rs
```diff
@@ -625,31 +627,9 @@ pub trait PushService {
...
-                bundle = bundle.signature(&signed_pre_key.signature);
-            }
-            if let Some(pre_key) = device.pre_key {
-                bundle = bundle.pre_key(
-                    pre_key.key_id,
-                    &PublicKey::decode_point(context, &pre_key.public_key)?,
-                );
-            }
-            pre_keys.push(bundle.build()?)
+            pre_keys.push(device.into_bundle(identity.clone())?);
```
rubdos, line 632:
```suggestion
            pre_keys.push(device.into_bundle(identity)?);
```

libsignal-service/src/sealed_session_cipher.rs
```diff
@@ -660,23 +686,30 @@ impl CertificateValidator {
...
+            .trust_root
             .verify_signature(
                 &server_certificate.certificate,
                 &server_certificate.signature,
             )
             .map_err(|e| {
                 error!("failed to verify server certificate: {}", e);
                 SealedSessionError::InvalidCertificate
-            })?;
+            })?
```
rubdos, line 698: clippy is annoyed here

libsignal-service/src/sealed_session_cipher.rs
```diff
@@ -660,23 +686,30 @@ impl CertificateValidator {
...
 
-        server_certificate
+        if !server_certificate
             .key
             .verify_signature(&certificate.certificate, &certificate.signature)
             .map_err(|e| {
                 error!("failed to verify certificate: {}", e);
                 SealedSessionError::InvalidCertificate
-            })?;
+            })?
```
rubdos, line 709: clippy again

libsignal-service/src/sender.rs
rubdos: What do we do with this, @gferon ?

libsignal-service/src/sender.rs
```diff
@@ -680,53 +692,57 @@ where
...
         recipient: &ServiceAddress,
         unidentified_access: Option<&UnidentifiedAccess>,
-        device_id: i32,
+        device_id: u32,
         content: &[u8],
     ) -> Result<OutgoingPushMessage, MessageSenderError> {
         let recipient_address = get_preferred_protocol_address(
-            &self.cipher.store_context,
-            recipient.clone(),
+            &mut self.session_store,
```
rubdos, line 699:
```suggestion
            &self.session_store,
```

libsignal-service/src/sender.rs
```diff
@@ -680,53 +692,57 @@ where
...
         log::trace!("encrypting message for {:?}", recipient_address);
 
         if !self
-            .cipher
-            .store_context
-            .contains_session(&recipient_address)?
+            .session_store
+            .load_session(&recipient_address, None)
+            .await?
+            .is_some()
```
rubdos, line 710: Suggestion feature on GH doesn't let me, but this should be `self.[..].is_none()` instead, apparently

libsignal-service/src/sender.rs
```diff
@@ -680,53 +692,57 @@ where
...
 
                 let pre_key_address = get_preferred_protocol_address(
-                    &self.cipher.store_context,
-                    recipient.clone(),
-                    pre_key_bundle.device_id(),
-                )?;
-                let session_builder = SessionBuilder::new(
-                    &self.cipher.context,
-                    &self.cipher.store_context,
+                    &mut self.session_store,
```
rubdos, line 724:
```suggestion
                    &self.session_store,
```

rubdos (commented):
Mostly clippy, there are still some unresolved discussions, @gferon.

libsignal-service/src/push_service.rs
gferon: Had a look at it and all of the fields, LGTM!

PR #85, opened 2021-04-23
Allow removing HTTP credentials using credentials_override

gferon:
Using an enum like this avoids us to have types like `Option<Option<HttpCredentials>>` that are hard to parse by humans.

This is also helpful for structs that perform authenticated and non-authenticated requests, which can now be a hidden implementation detail.

### DIFF
--- a/libsignal-service/src/groups_v2/manager.rs
+++ b/libsignal-service/src/groups_v2/manager.rs
@@ -7,7 +7,7 @@ use crate::{
     configuration::Endpoint,
     prelude::{PushService, ServiceError},
     proto::DecryptedGroup,
-    push_service::HttpCredentials,
+    push_service::{HttpAuth, HttpAuthOverride},
 };
 
 use rand::RngCore;
@@ -125,7 +125,7 @@ impl<'a, S: PushService, C: CredentialsCache> GroupsManager<'a, S, C> {
         &mut self,
         uuid: Uuid,
         group_secret_params: GroupSecretParams,
-    ) -> Result<HttpCredentials, ServiceError> {
+    ) -> Result<HttpAuth, ServiceError> {
         let today = Self::current_time_days();
         let auth_credential_response = if let Some(auth_credential_response) =
             self.credentials_cache.get(&today)?
@@ -162,7 +162,7 @@ impl<'a, S: PushService, C: CredentialsCache> GroupsManager<'a, S, C> {
             format!("/v1/certificate/group/{}/{}", today, today_plus_7_days);
 
         self.push_service
-            .get_json(Endpoint::Service, &path, None)
+            .get_json(Endpoint::Service, &path, HttpAuthOverride::NoOverride)
             .await
     }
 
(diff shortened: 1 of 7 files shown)


libsignal-service/src/push_service.rs
```diff
@@ -119,9 +119,9 @@ pub struct PreKeyStatus {
 }
 
 #[derive(Clone)]
-pub struct HttpCredentials {
-    pub username: String,
-    pub password: String,
+pub enum HttpAuth {
```
rubdos, line 122: May I suggest a manual implementation of `Debug` here, that prints *only the username* (and possibly "password: [..]" or something like that)?

libsignal-service/src/groups_v2/manager.rs
```diff
@@ -211,13 +211,13 @@ impl<'a, S: PushService, C: CredentialsCache> GroupsManager<'a, S, C> {
...
 
-        Ok(HttpCredentials { username, password })
+        Ok(HttpAuth::Basic { username, password })
     }
 
     pub async fn get_group(
         &mut self,
         group_secret_params: GroupSecretParams,
-        credentials: HttpCredentials,
+        credentials: HttpAuth,
```
rubdos, line 220: What's the semantics of setting this to `HttpAuth::None`?

gferon: see for example [the first use of it here](https://github.com/Michael-F-Bryan/libsignal-service-rs/pull/85/files#diff-78c1e3016616476335a7b4df6a1171673c2af645b0740309febf03a96f35f95aR367)

I suspect that it'll also be useful for structs that are initialized without credentials and need them later (mostly when linking a secondary device for example).

rubdos:
So, we're modeling tri-state here, right?

1. Do nothing
2. Override default credentials and set *no credentials*
3. Override default credentials and set *these credentials*.

The problem, I think, is that certain functions (notably `put_json` et al) require this tri-state argument, while other methods (e.g. related to the group stuff) require a filed credential. Is it, maybe, an option to have something like

```rust
enum HttpAuthOverride {
    NoOverride,
    OverrideWith(Option<HttpAuth>),
}

struct HttpAuth {}
```

... possibly you can model the full tri-state in `HttpAuthOverride`, getting rid of the `Option`:

```rust
enum HttpAuthOverride {
    NoOverride,
    Unidentified,
    Identified(HttpAuth),
}
```

libsignal-service/src/groups_v2/manager.rs
rubdos: I mean the semantics in the `get_group` specifically.

gferon:
I see, I missed the `get_group` use somehow, which is rather a basic one 😄 I'll try with this suggestion and see how it goes.

libsignal-service/src/push_service.rs
```diff
@@ -112,25 +112,38 @@ pub struct DeviceCapabilities {
...
     pub password: String,
 }
 
-impl ProfileKey {
-    pub fn derive_access_key(&self) -> Vec<u8> {
-        let key = GenericArray::from_slice(&self.0);
-        let cipher = Aes256Gcm::new(key);
-        let nonce = GenericArray::from_slice(&[0u8; 12]);
-        let buf = [0u8; 16];
-        cipher.encrypt(nonce, &buf[..]).unwrap()
```
rubdos, line 133: Something weird happened to the diff here, mind fixing that up?

rubdos (changes requested):
I like the feel of this, good job.  You like it too?

Small remark wrt. the diff, but it probably doesn't matter. Feel free to decide on that question and merge.

libsignal-service/src/push_service.rs
gferon: Done!

PR #86, opened 2021-04-23
Fetch/decode contacts from sync messages

gferon:


libsignal-service/src/models.rs
```diff
@@ -1,33 +1,75 @@
...
-    pub members: Vec<String>,
-    pub avatar: Option<Vec<u8>>,
+    pub color: Option<String>,
+    pub verified: Verified,
+    pub profile_key: Vec<u8>,
+    pub blocked: bool,
+    pub expire_timer: u32,
+    pub inbox_position: u32,
+    pub archived: bool,
+    pub avatar: Option<Attachment<Bytes>>,
```
rubdos, line 27: I wonder whether it makes sense to really store this as bytes, instead of a pointer to where the attachment is stored?

libsignal-service/src/receiver.rs
```diff
@@ -18,6 +24,8 @@ pub enum MessageReceiverError {
 }
 
 impl<Service: PushService> MessageReceiver<Service> {
+    // TODO: to avoid providing the wrong service/wrong credentials
+    // change it like LinkingManager or ProvisioningManager
```
rubdos, line 28: `TODO` marker (either make an issue or resolve here)

rubdos (commented):
I like the Iterator approach! Maybe it makes sense to make it a `Stream` instead, which downloads the attachments as you go?

libsignal-service/src/models.rs
gferon: well, if you look at the `Attachment` struct, it can be anything that implements `Reader` - in this particular context (contacts attachment from a sync message), avatars are always binary data you get directly without the need to fetch it.

gferon:
> I like the Iterator approach! Maybe it makes sense to make it a `Stream` instead, which downloads the attachments as you go?

There's really only one download here, and we don't have a way (as far as I know) to decrypt a `Stream` and `prost::decode` it as we go?

libsignal-service/src/receiver.rs
```diff
@@ -51,4 +59,89 @@ impl<Service: PushService> MessageReceiver<Service> {
...
+        if let Some(ref blob) = contacts.blob {
+            use futures::io::AsyncReadExt;
+
+            let mut stream = loop {
+                let r = self.service.get_attachment(&blob).await;
+                match r {
+                    Ok(stream) => break stream,
+                    Err(ServiceError::Timeout { .. }) => {
+                        log::warn!("get_attachment timed out, retrying")
+                    }
```
rubdos, line 79: I wonder whether we should "fail after five times" or something with a "try again" error? Usually, these time-outs require some backing off, no? What does the NodeJS impl do with this?

rubdos:
> > I like the Iterator approach! Maybe it makes sense to make it a `Stream` instead, which downloads the attachments as you go?
> 
> There's really only one download here, and we don't have a way (as far as I know) to decrypt a `Stream` and `prost::decode` it as we go?

Ah, I misread the code there. I saw a loop and thought it downloaded the avatar for every contact. No worries then :-)

libsignal-service/src/receiver.rs
gferon: I just baked in a super basic "retry in a loop" - I didn't go the exponential backoff ways because: 1. it would tie `libsignal-service` to a Futures executor flavor (since you have to sleep in async), 2. our error reporting is not granular enough to handle errors that are worth retrying that aren't timeouts

rubdos (approved):
:+1:

libsignal-service/src/receiver.rs
rubdos: Perfect. Caller should do exp-backoff if needed then. I just imagined running this loop on a disconnected phone that for some reason calls it "time out" (happens on e.g. 4G). You'd just have a near-100%-CPU loop then. This is better :-)

PR #87, opened 2021-04-24
Retrieve profile by uuid.

boxdot:
We only retrieve `name`, `about` and `about_emoji`.

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -286,7 +286,7 @@ impl<Service: PushService> AccountManager<Service> {
     pub async fn retrieve_profile(
         &mut self,
         uuid: uuid::Uuid,
-    ) -> Result<Profile, ServiceError> {
+    ) -> Result<Profile, ProfileManagerError> {
         let profile_key =
             self.profile_key.expect("set profile key in AccountManager");
         let profile_key = ProfileKey::create(profile_key);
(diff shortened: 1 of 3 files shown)


libsignal-service/src/account_manager.rs
```diff
@@ -276,6 +283,38 @@ impl<Service: PushService> AccountManager<Service> {
...
+
+        let encrypted_profile = self
+            .service
+            .retrieve_profile_by_id(&uuid.to_string())
+            .await?;
+
+        // Profile decryption
+        let name = encrypted_profile
+            .name
+            .and_then(|data| profile_cipher.decrypt_name(data).ok()?);
```
rubdos, line 303: You're dismissing the error here. I suggest you add `ProfileCipherError` to the `ServiceError`, and then use something like

```suggestion
            .map(|data| profile_cipher.decrypt_name(data)).transpose()?;
```

(same for the other fields)

rubdos (changes requested):
Error handling is not on point yet, if you fix that up I'll be glad to merge.

libsignal-service/src/account_manager.rs
boxdot: Thanks. I misread `upload_versioned_profile`. I thought the errors are dismissed there, but they are not.

rubdos (approved):
Looks nice like this :-)

PR #88, opened 2021-04-26
Expose ProfileManagerError on public interface.

boxdot:


rubdos:
Makes sense, I suppose :-)

boxdot:
> Makes sense, I suppose :-)

Yeah, it is needed in `presage`.

PR #93, opened 2021-04-30
Reuse http client.

boxdot:
The push service is provided to managers from outside, instead of
being created inside. This makes it possible to reuse the same
push service (http client) between managers.

### DIFF
--- a/libsignal-service/src/provisioning/manager.rs
+++ b/libsignal-service/src/provisioning/manager.rs
@@ -11,8 +11,7 @@ use super::{
 };
 
 use crate::{
-    configuration::{Endpoint, ServiceConfiguration, SignalingKey},
-    messagepipe::ServiceCredentials,
+    configuration::{Endpoint, ServiceCredentials, SignalingKey},
     push_service::{
         DeviceCapabilities, DeviceId, HttpAuthOverride, PushService,
         ServiceError,
@@ -104,11 +103,12 @@ pub enum VerificationCodeResponse {
     Issued,
 }
 
-#[derive(Clone)]
-pub struct ProvisioningManager<P: PushService> {
-    push_service: P,
+pub struct ProvisioningManager<'a, P: PushService + 'a> {
+    push_service: &'a mut P,
     phone_number: PhoneNumber,
+    password: String,
 }
+
 pub enum SecondaryDeviceProvisioning {
     Url(Url),
     NewDeviceRegistration {
@@ -122,26 +122,16 @@ pub enum SecondaryDeviceProvisioning {
     },
 }
 
-impl<P: PushService> ProvisioningManager<P> {
+impl<'a, P: PushService + 'a> ProvisioningManager<'a, P> {
     pub fn new(
-        cfg: impl Into<ServiceConfiguration>,
-        user_agent: String,
+        push_service: &'a mut P,
         phone_number: PhoneNumber,
         password: String,
     ) -> Self {
         Self {
-            phone_number: phone_number.clone(),
-            push_service: P::new(
-                cfg,
-                Some(ServiceCredentials {
-                    phonenumber: phone_number,
-                    password: Some(password),
-                    uuid: None,
-                    signaling_key: None,
-                    device_id: None,
-                }),
-                user_agent,
-            ),
+            push_service,
+            phone_number,
+            password,
         }
     }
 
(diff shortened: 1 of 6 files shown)


libsignal-service/src/provisioning/manager.rs
```diff
@@ -264,23 +261,16 @@ impl<P: PushService> ProvisioningManager<P> {
 
 #[derive(Clone)]
 pub struct LinkingManager<P: PushService> {
-    cfg: ServiceConfiguration,
-    user_agent: String,
-    password: String,
     push_service: P,
+    // used in credentials forwarded to the `ProvisioningManager`
+    password: Arc<String>,
```
rubdos, line 266: What is the benefit of Arc'ing this?

libsignal-service/src/push_service.rs
```diff
@@ -292,15 +292,11 @@ pub enum ServiceError {
 }
 
 #[async_trait::async_trait(?Send)]
-pub trait PushService {
+pub trait PushService: Clone {
```
rubdos, line 295: I'm not sure whether I like to really *enforce* `: Clone`.  This is only used in the provisioning manager.  What about passing the provisioning manager the `service` by (mutable) reference? Can't we generically `impl PushService for &mut PushService` or something like that?

rubdos: If you still want to go the `: Clone` route, maybe just put the trait bound on the provisioningmanager impl block, or even on that single method.

rubdos (commented):
Some comments. Also, this conflicted when I merged #83, sorry for that.

libsignal-service/src/provisioning/manager.rs
boxdot: Optimization for light-weight cloning.

libsignal-service/src/push_service.rs
boxdot: Let me try the `&mut` approach.

libsignal-service/src/provisioning/manager.rs
rubdos: Sounds like premature optimization to me. `Arc` comes with its own overhead on every access, compared to a single malloc on clone (and clone happens rarely). Wouldn't do any of this without a clear indication for a bottleneck of some sort, and if there is a real bottleneck, I would use `password: &'a str` or `password: impl AsRef<str>` instead.

This just reads like poorly Java-ported code.

boxdot: I'll remove it for now and come back with a better argument when optimizing for clones (which happens on each incoming message).

libsignal-service/src/push_service.rs
boxdot: It worked. Please see the rebased review.

rubdos (approved):
Still a single comment. If you disagree, let me know why and I'll merge this.

Aside: it looks like secondary device provisioning still assumes we have phone numbers. Signal probably stopped assuming this upstream, we might want to make an issue about that pretty soon.

libsignal-service/src/provisioning/manager.rs
```diff
@@ -255,27 +245,35 @@ impl<P: PushService> ProvisioningManager<P> {
...
+            password: Some(self.password.clone()),
+            signaling_key: None,
+            device_id: None,
+        };
+        if let Some(auth) = credentials.authorization() {
+            HttpAuthOverride::Identified(auth)
+        } else {
+            HttpAuthOverride::NoOverride
+        }
+    }
```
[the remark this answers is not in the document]
boxdot: This might be a little bit surprising that after using ProvisitionManager the auth has changed. This means the client of provisioning would need to save the http auth to be able to restore it.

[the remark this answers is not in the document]
rubdos: I get what you're saying. On the other hand, our `PushService` is owned by the `ProvisionManager` and never returned, *and* all the `PushService` implementations clone their `Credentials`, and thus don't have shared mutability for them. This shouldn't be observable for the outside world as far as I can tell.

This does put a constraint on `PushService` implementors though, which we probably want to document. Given this new information, would you change your mind or do I merge as-is? Either is good to me.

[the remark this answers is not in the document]
rubdos: > Hmm, the PushService is not owned by ProvisionManager anymore since we give a &mut instance to it.

Makes a lot of sense to set overrides then. 👍️

PR #97, opened 2021-05-02
Add serde modules for public/private/signaling keys

gferon:
This helps with serializing a struct containing all the registration data easily (like in `presage`).


libsignal-service/src/push_service.rs
```diff
@@ -147,6 +148,42 @@ impl ProfileKey {
...
+                .try_into()
+                .map_err(|buf: Vec<u8>| {
+                    serde::de::Error::invalid_length(
+                        buf.len(),
+                        &"invalid profile key length",
+                    )
+                })?,
+        ))
+    }
+}
```
rubdos, line 185: I can't imagine that this isn't possible with the `with_base64` module that we already have and maybe a `#[serde(flatten)]`?

gferon: Our `serde_base64` will not deserialize to an array of known size, but to a `Vec<u8>` so even though `Serialize` can be derived, `Deserialize` can't.

rubdos: Ah right, and for a module to do so, you probably need HKT. Dismissed!

rubdos:
LGTM

PR #98, opened 2021-05-03
Replace deprecated aes-ctr crate.

boxdot:


gferon (approved):
Looks good to me, I'd just like to ask what @rubdos think about it before merging.

Issue #75, opened 2021-04-05
send_message_to_group assumes that recipients[0] is our own account


rubdos:
Either we document this necessity, or we change the method to be clear about this.

gferon:
This is not the case anymore following the merge of #103 😄

Issue #25, opened 2020-10-28
Send messages through MessagePipe


rubdos:

gferon:
Done in #170

PR #101, opened 2021-05-06
Send messages using the (websocket) message pipe

gferon:
This adds the ability to submit messages to the `MessagePipe` (resolving #25).

Included is a new wrapper struct called `MessageToSend` which is deliberately not named `Message` as I thought it was too generic, any other name suggestion is appreciated.

From a high-level API view, it should look like:
```rust
let message_pipe = MessagePipe::from_socket(websocket, stream, credentials, message_sender);

// sender implements `Sink` and receiver implements `Stream` 
let (sender, receiver) = message_pipe.split();
```

TODO:
- [ ] Finish factoring the code to avoid duplication between `sender.rs` and `messagepipe.rs` _or_ remove sync message sending abilities altogether? Paging Dr. @rubdos
- [ ] Figure out how group messages should be sent (related to #75):
  1. the caller is responsible
  2. provide a helper method, like the one we have currently (`send_message_to_group`)
  3. something else
- [ ] Figure out if/why messages arrive only on the primary device, per @boxdot report when testing
- [ ] Maybe figure out the best way to have the `MessagePipe` work without having to poll on the receiving stream in order for sending to work... I suppose the obvious reflex would be to use `tokio::spawn`, which is not really available here. From a pure API point of view, we might want to `impl Future` on `MessagePipe` but then the user would be responsible to poll on 3 futures in order for the whole thing to work. I'm torn.

Should help @boxdot with https://github.com/boxdot/gurk-rs/issues/50

### DIFF
--- a/libsignal-service/src/messagepipe.rs
+++ b/libsignal-service/src/messagepipe.rs
@@ -30,8 +30,8 @@ use crate::{
     push_service::{PushService, ServiceError},
     sealed_session_cipher::UnidentifiedAccess,
     sender::{
-        create_multi_device_sent_transcript_content, MessageToSend,
-        OutgoingPushMessages, SendMessageResponse, SentMessage,
+        create_multi_device_sent_transcript_content, OutgoingPushMessages,
+        SendMessageResponse, SentMessage,
     },
     session_store::SessionStoreExt,
     ServiceAddress,
@@ -40,6 +40,14 @@ use crate::{
 pub enum WebSocketStreamItem {
     Message(Bytes),
     KeepAliveRequest,
+}
+
+pub struct MessageToSend {
+    pub recipients: Vec<ServiceAddress>,
+    pub unidentified_access: Option<UnidentifiedAccess>,
+    pub content_body: ContentBody,
+    pub timestamp: u64,
+    pub online: bool,
 }
 
 #[async_trait::async_trait(?Send)]
@@ -145,7 +153,8 @@ where
         }
     }
 
-    /// Worker task that
+    /// Worker task that runs the message pipe loop which concurrently looks for incoming message to process
+    /// and processes frames received in the web-socket
     async fn run(
         mut self,
         mut outgoing_sender: Sender<Result<Envelope, ServiceError>>,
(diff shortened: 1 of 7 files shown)


libsignal-service/src/messagepipe.rs
```diff
@@ -119,9 +146,11 @@ impl<WS: WebSocketService> MessagePipe<WS> {
     }
 
     /// Worker task that
-    async fn run(
+    async fn run<'a>(
```
rubdos, line 149: Unused lifetime?

Be-ing:
@nanuc uses a WebSocket to communicate between the Go backend of Axolotl and the Vue.js frontend. Might this be relevant to potentially using libsignal-service-rs in Axolotl?

rubdos:
> @nanu-c uses a WebSocket to communicate between the Go backend of Axolotl and the Vue.js frontend. Might this be relevant to potentially using libsignal-service-rs in Axolotl?

No, this is using the websocket between the Signal servers and the client. Sounds like Axolotl uses a separate websocket between the client and the UI.

rubdos:
> This adds the ability to submit messages to the `MessagePipe` (resolving #25).
> 
> Included is a new wrapper struct called `MessageToSend` which is deliberately not named `Message` as I thought it was too generic, any other name suggestion is appreciated.

`QueuedMessage`? That would imply that the `MessagePipe` implements some kind of smart-ish queueing system though.

>     * [ ]  Finish factoring the code to avoid duplication between `sender.rs` and `messagepipe.rs` _or_ remove sync message sending abilities altogether? Paging Dr. @rubdos

Nah, wouldn't drop it. Not everyone wants to use the full-fledged pipe I think. Maybe we can do some smart traits and wrappers to abstract over both?

>     * [ ]  Figure out how group messages should be sent (related to #75):
>       
>       1. the caller is responsible
>       2. provide a helper method, like the one we have currently (`send_message_to_group`)
>       3. something else

I would keep the sender and pipe in sync as much as possible, so possibly an abstraction over both solves this too.

>     * [ ]  Maybe figure out the best way to have the `MessagePipe` work without having to poll on the receiving stream in order for sending to work... I suppose the obvious reflex would be to use `tokio::spawn`, which is not really available here. From a pure API point of view, we might want to `impl Future` on `MessagePipe` but then the user would be responsible to poll on 3 futures in order for the whole thing to work. I'm torn.

`MessagePipe` already implements `Stream`, so you can poll whatever you need in there, no?

rubdos:
> > MessagePipe already implements Stream, so you can poll whatever you need in there, no?
> 
> The problem I see here, is that somebody uses `.split()` and then submits a message in the `Sender` without polling on the `Stream` and nothing will happen. I don't think it's a massive problem though, just feels like it'd be easy to miss.

Well, those who don't use the stream should use the bare sender, no? You can also implement it in the `Sink` part, of course.

PR #103, opened 2021-05-08
Only send one sync message when sending group messages

gferon:
and properly populate the UnidentifiedDeliveryStatus protobuf message

Sending multiple sync messages led to some clients showing the
same message multiple times and was a mistake in the first place.

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -47,8 +47,10 @@ pub struct SendMessageResponse {
     pub needs_sync: bool,
 }
 
+pub type SendMessageResult = Result<SentMessage, MessageSenderError>;
+
 #[derive(Debug, Clone)]
-pub struct SendMessageResult {
+pub struct SentMessage {
     recipient: ServiceAddress,
     unidentified: bool,
     needs_sync: bool,
@@ -314,7 +316,7 @@ where
         message: impl Into<ContentBody>,
         timestamp: u64,
         online: bool,
-    ) -> Result<SendMessageResult, MessageSenderError> {
+    ) -> SendMessageResult {
         let content_body = message.into();
 
         use crate::proto::data_message::Flags;
@@ -339,7 +341,7 @@ where
             // if we sent a data message and we have linked devices, we need to send a sync message
             (
                 ContentBody::DataMessage(message),
-                Ok(SendMessageResult { needs_sync, .. }),
+                Ok(SentMessage { needs_sync, .. }),
             ) if *needs_sync => {
                 log::debug!("sending multi-device sync message");
                 let sync_message = self
@@ -347,6 +349,7 @@ where
                         Some(&recipient),
                         Some(message.clone()),
                         timestamp,
+                        &[&result],
                     );
                 self.try_send_message(
                     (&self.local_address).clone(),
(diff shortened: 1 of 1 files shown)


libsignal-service/src/sender.rs
```diff
@@ -750,7 +751,25 @@ where
         recipient: Option<&ServiceAddress>,
         data_message: Option<crate::proto::DataMessage>,
         timestamp: u64,
+        send_message_results: &[&SendMessageResult],
```
rubdos, line 754: If you make this `&[SendMessageResult]`, you can just borrow the `Vec` above instead of creating a new vec with only references. The extra `&` inside the slide doesn't really give you anything extra useful in terms of reference.

libsignal-service/src/sender.rs
```diff
@@ -383,56 +386,54 @@ where
...
-                result => result,
+                _ => (),
             };
+
+            results.push(result);
+        }
+
+        // we only need to send a synchronization message once
+        if needs_sync_in_results {
+            let results_ref: Vec<&SendMessageResult> = results.iter().collect();
```
rubdos, line 418: See comment below.

libsignal-service/src/sender.rs
```diff
@@ -750,7 +751,25 @@ where
...
+                        recipient,
+                        unidentified,
+                        ..
+                    }) => Some(UnidentifiedDeliveryStatus {
+                        destination_e164: recipient.e164(),
+                        destination_uuid: recipient.uuid.map(|s| s.to_string()),
+                        unidentified: Some(*unidentified),
+                    }),
+                    _ => None,
+                })
```
rubdos, line 771: I'm thinking that you may be able to get rid of the `match` here:

```suggestion
                .filter_map(Result::ok)
                .map(|sent| {
                    let SentMessage { recipient, unidentified, .. } = sent;
                    UnidentifiedDeliveryStatus {
                        destination_e164: recipient.e164(),
                        destination_uuid: recipient.uuid.map(|s| s.to_string()),
                        unidentified: Some(*unidentified),
                    }
                })
```

Not sure whether it's cleaner or not, and not sure whether it compiles as-is. It says "filter on Ok, then map those to UnidentifiedDeliveryStatus". I think I like it ;p

rubdos (commented):
Looks good to me in general, just some nitpicks.

libsignal-service/src/sender.rs
gferon: `Result::ok` won't work, but `|result| result.ok()` will because of type inference issues. Doing it now.

rubdos: Not even if you annotate the type?

gferon: First of all, yes, but why when the compiler can figure it out for you 🔮 I'm not even sure we can do it as nicely, since we're working on a slice, and [need to filter on a reference anyway](https://github.com/whisperfish/libsignal-service-rs/pull/103/commits/26a3665834e26b9340ebf3c5b32089cf7328d2c1#diff-9ae521e6fffe43a399d2d3007bbccc81a73e245f5aa5663b007af1fa97dbcc26L760).

rubdos: I don't see why it needs to be a reference, but if you say it has to be, then it's alright with me now.

rubdos:
If you're satisfied with this, feel free to squash-and-merge!

PR #104, opened 2021-05-18
Refactor delete_service_addr_device_session, catch non-existing session on deletion

rubdos:
We had `.unwrap()` in Whisperfish on `delete_session`. Since `delete_session` was called on both e164 and uuid sessions, this unconditionally causes a panic.

I have altered Whisperfish (https://gitlab.com/whisperfish/whisperfish/-/commit/246869a7ab293b79aa5299fb53ad7bebbd2688a0) to return `SignalProtocolError::SessionNotFound`, and this pull request makes libsignal-service understand that that's an error to be expected.

Meanwhile, I take the opportunity to refactor the deletion logic, since it was copy-pasted a few times. A better suggestion for that method name is appreciated.

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -484,22 +484,12 @@ where
                             "dropping session with device {}",
                             extra_device_id
                         );
-                        if let Some(uuid) = recipient.uuid {
-                            self.session_store
-                                .delete_session(&ProtocolAddress::new(
-                                    uuid.to_string(),
-                                    *extra_device_id,
-                                ))
-                                .await?;
-                        }
-                        if let Some(e164) = recipient.e164() {
-                            self.session_store
-                                .delete_session(&ProtocolAddress::new(
-                                    e164,
-                                    *extra_device_id,
-                                ))
-                                .await?;
-                        }
+                        self.session_store
+                            .delete_service_addr_device_session(
+                                &recipient,
+                                *extra_device_id,
+                            )
+                            .await?;
                     }
 
                     for missing_device_id in &m.missing_devices {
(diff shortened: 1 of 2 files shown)


libsignal-service/src/session_store.rs
```diff
@@ -26,4 +28,42 @@ pub trait SessionStoreExt: SessionStore {
         &self,
         address: &str,
     ) -> Result<usize, SignalProtocolError>;
+
+    /// Remove a session record for a recipient ID + device ID tuple.
+    async fn delete_service_addr_device_session(
```
gferon, line 33: The implementation looks good to me, but I have the following questions:

1. since we probably don't want users to implement this method themselves, what about having it as a method of `ServiceAddress` instead? We already have `get_sub_device_sessions` there.
2. I also not sure if we should use `ServiceAddress` in those traits, since we might want to try and merge them back to `libsignal-protocol`?

libsignal-service/src/sender.rs
```diff
@@ -540,22 +530,12 @@ where
...
-                        }
-                        if let Some(e164) = recipient.e164() {
-                            self.session_store
-                                .delete_session(&ProtocolAddress::new(
-                                    e164,
-                                    *extra_device_id,
-                                ))
-                                .await?;
-                        }
+                        self.session_store
```
gferon, line 533: much better to not duplicate this logic everywhere!

libsignal-service/src/session_store.rs
rubdos: > I also not sure if we should use `ServiceAddress` in those traits, since we might want to try and merge them back to `libsignal-protocol`?

You would rather pass `uuid`, `e164` and `device_id` separately? IMO, this is a feature of the session store, libsignal-protocol only cares about one session at a time.

> since we probably don't want users to implement this method themselves, what about having it as a method of `ServiceAddress` instead? We already have `get_sub_device_sessions` there.

Hmm, I get why you say to keep this method non-overridable; maybe it shouldn't be part of the public API indeed. OTOH, moving it into ServiceAddress would mean that you'd call `service_address.delete_sessions_for_device(&mut some_session_store, device_id)`, which IMO sounds even worse.

What I could maybe get behind, is a `pub(crate) trait SessionStoreExtPrivate: SessionStore {}` which implements these kind of methods.

rubdos: Argument *pro* having this method overridable: an SQL back-end could implement this as in `DELETE FROM sessions WHERE uuid={} OR e164={}`.

I think we should just merge as-is, no?

gferon: Sure, go ahead!

PR #105, opened 2021-05-22
Support Send for libsignal-service futures

lushing:
This is my first pass at adding some support for `Send` within the `libsignal-service` futures. There will likely be other changes required for full compatibility (including upstream changes to `libsignal-client`), but these may be best left for future PRs.

The solution here still feels a bit odd, but it seemed like the best solution to avoid a significant amount of code duplication or boilerplate. I definitely welcome any suggestions/improvements!

### DIFF
--- a/libsignal-service/Cargo.toml
+++ b/libsignal-service/Cargo.toml
@@ -43,4 +43,4 @@ tokio = { version = "1.0", features = [ "macros" ] }
 
 [features]
 prefer-e164 = []
-require-send = []
+unsend-futures = []
--- a/libsignal-service/Cargo.toml
+++ b/libsignal-service/Cargo.toml
@@ -43,3 +43,4 @@ tokio = { version = "1.0", features = [ "macros" ] }
 
 [features]
 prefer-e164 = []
+require-send = []
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -19,7 +19,6 @@ use crate::{
         AccountAttributes, DeviceCapabilities, HttpAuthOverride, PushService,
         ServiceError,
     },
-    MaybeSend,
 };
 
 pub struct AccountManager<Service> {
@@ -60,7 +59,7 @@ const PRE_KEY_MINIMUM: u32 = 10;
 const PRE_KEY_BATCH_SIZE: u32 = 100;
 const PRE_KEY_MEDIUM_MAX_VALUE: u32 = 0xFFFFFF;
 
-impl<Service: PushService + MaybeSend> AccountManager<Service> {
+impl<Service: PushService> AccountManager<Service> {
     pub fn new(service: Service, profile_key: Option<[u8; 32]>) -> Self {
         Self {
             service,
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -19,6 +19,7 @@ use crate::{
         AccountAttributes, DeviceCapabilities, HttpAuthOverride, PushService,
         ServiceError,
     },
+    MaybeSend,
 };
 
 pub struct AccountManager<Service> {
@@ -59,7 +60,7 @@ const PRE_KEY_MINIMUM: u32 = 10;
 const PRE_KEY_BATCH_SIZE: u32 = 100;
 const PRE_KEY_MEDIUM_MAX_VALUE: u32 = 0xFFFFFF;
 
-impl<Service: PushService> AccountManager<Service> {
+impl<Service: PushService + MaybeSend> AccountManager<Service> {
     pub fn new(service: Service, profile_key: Option<[u8; 32]>) -> Self {
         Self {
             service,
--- a/libsignal-service/src/lib.rs
+++ b/libsignal-service/src/lib.rs
@@ -40,6 +40,15 @@ pub const GROUP_UPDATE_FLAG: u32 = 1;
 /// GROUP_LEAVE_FLAG signals that this message is a group leave message.
 pub const GROUP_LEAVE_FLAG: u32 = 2;
 
+/// This trait allows for the conditional support of Send compatible futures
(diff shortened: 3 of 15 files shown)


libsignal-service/src/lib.rs
```diff
@@ -40,6 +40,16 @@ pub const GROUP_UPDATE_FLAG: u32 = 1;
...
+#[cfg(feature = "require-send")]
+pub trait MaybeSend: Send {}
+#[cfg(feature = "require-send")]
+impl<T> MaybeSend for T where T: Send {}
+
+#[cfg(not(feature = "require-send"))]
+pub trait MaybeSend {}
+#[cfg(not(feature = "require-send"))]
+impl<T> MaybeSend for T {}
+
```
rubdos, line 52: I've found https://github.com/rust-lang/rust/issues/41517 to be interesting.

libsignal-service/src/account_manager.rs
```diff
@@ -59,7 +60,7 @@ const PRE_KEY_MINIMUM: u32 = 10;
 const PRE_KEY_BATCH_SIZE: u32 = 100;
 const PRE_KEY_MEDIUM_MAX_VALUE: u32 = 0xFFFFFF;
 
-impl<Service: PushService> AccountManager<Service> {
+impl<Service: PushService + MaybeSend> AccountManager<Service> {
```
rubdos, line 63: I wonder, can't we pull this in the definition of PushService itself? Declare `trait PushService: Send` with the feature, and `trait PushService` without it?

lushing: Absolutely! Can't believe it didn't cross my mind, this way is much simpler.

libsignal-service/src/lib.rs
lushing: I had come across this while trying to figure out that bit, but does libsignal-service intend on using nightly?

gferon:
Thanks a lot! I'll try to have a look tomorrow as well. I suppose you tried this with https://github.com/signalapp/libsignal-client/pull/297 ?

lushing:
@gferon Haven't yet tested it with your branch, but everything here compiles as-is and can likely be merged without any additional changes or custom dependencies. I think it's possible to achieve some degree of `Send` compatibility even without the upstream changes to `libsignal-client`, but full `Send` support will require that your PR is merged.

libsignal-service/src/push_service.rs
```diff
@@ -354,7 +354,7 @@ pub enum ServiceError {
 
 #[cfg_attr(feature = "require-send", async_trait::async_trait)]
 #[cfg_attr(not(feature = "require-send"), async_trait::async_trait(?Send))]
-pub trait PushService {
+pub trait PushService: MaybeSend {
```
rubdos, line 357: Hmm, this still has the awkward `MaybeSend`, but I suppose that without the trait-aliases it's going to be a lot of code copy-paste to get that out.

libsignal-service/src/lib.rs
rubdos: Oh, no, not at all. Just leaving it for reference!

libsignal-service/src/push_service.rs
lushing: Indeed, I don't think there is any other way to conditionally require `Send` while staying on stable. It's either the weird trait, or you have to copy paste the entire `pub trait PushService { /* ... */ }`.

rubdos: You could probably macro it out, if you're up to that, but I'd wait with that until @gferon gives some fb :)

gferon:
I had a look and I think it's the most elegant solution we can come up with (at least for now!)

Could you maybe add a block comment somewhere explaining why this is necessary? It could mention https://github.com/actix/actix-web/issues/1351 for example (to justify the difference between hyper and actix).

lushing:
sure thing @gferon added a few comments

gferon (approved):
LGTM from my side. @rubdos do you want to use a macro or not? I tend to prefer not to, since it can make it a little bit harder to read, but that's your call!

libsignal-service/Cargo.toml
```diff
@@ -43,3 +43,4 @@ tokio = { version = "1.0", features = [ "macros" ] }
 
 [features]
 prefer-e164 = []
+require-send = []
```
rubdos, line 46: I'm not sure whether `require-send` is the most appropriate name for the feature.

Additionally, it may be useful to document this in the README somehow.

libsignal-service/src/push_service.rs
rubdos: Let's not macro it. We might want a tracking issue for when trait aliases stabilize and are stable for ~6 months.

rubdos (commented):
Another thing I am wondering about: wouldn't it be more logical to have `Send` as opt-out? Instead of `require-send`, make it something like `unsend-futures`, because any implementation will be compatible with that feature, and then the Actix backend can opt-out of Send (instead of having Hyper to opt-in).

That should solve the CI failure too.

lushing:
@rubdos Hey, switched the name to `unsend-futures`, made it opt-out by default, and add a feature flag section to the README.

@gferon It seems the CI needs to be changed to accommodate for this weird workspace. Normally, feature flags are additive. However, in this case, the *presence* of a feature flag can actually break a crate. As one of the three crates has the `unsend-futures` flag enabled by default (`libsignal-service-actix`), cargo seems to enable this feature flag for all crates that have a `libsignal-service` dependency, including `libsignal-service-hyper`, which breaks the build.

This means that a top level `cargo build` will fail, but individually building each crate alone will succeed. You can refer to the recent Matrix conversation for more context, but it seems there is no way around this problem other than changing the CI to selectively compile things in a certain way. As it appears you are more acquainted with the CI, I'll leave such decisions to you.

rubdos:
@gferon please go ahead!

PR #108, opened 2021-08-02
Fix sending a sync message to the same device twice.

boxdot:
This happens when the session store contains overlapping sessions for
the uuid and the e164 phone number of the same account. A device
which has both session receives and shows a sync message twice.

### DIFF
--- a/libsignal-service/src/service_address.rs
+++ b/libsignal-service/src/service_address.rs
@@ -47,6 +47,8 @@ impl ServiceAddress {
             sub_device_sessions
                 .extend(session_store.get_sub_device_sessions(&e164).await?);
         }
+        sub_device_sessions.sort_unstable();
+        sub_device_sessions.dedup();
         Ok(sub_device_sessions)
     }
 }


rubdos:
You should *never* have overlapping sessions. The remote end cannot distinguish them, and this leads to undecryptable messages.

You should instead have logic [like this `merge_and_fetch_recipient`](https://gitlab.com/whisperfish/whisperfish/-/blob/master/src/store/mod.rs#L814), which decides when a phone number is moved or tied to a UUID. Better yet, Android's `RecipientDatabase::getAndPossiblyMerge`. An e164 and UUID that are "together" should never have separate sessions.

rubdos:
@boxdot , are you okay with closing this? Maybe we should document these requirements somehow, but it'd quite a challenge to document how to "correctly" implement the whole of Signal.

Maybe another approach here could be to give an error when overlapping sessions are retrieved, à la "This is a bug in the application and should be reported. The session with XYZ is now probably corrupt.".

boxdot:
I still don't have a working solution in my client. But I will close it.

rubdos:
> I still don't have a working solution in my client. But I will close it.

If you want, you can open an issue about session/uuid/e164 management and merging; it should be relatively easy to port the (currently incomplete) session merging logic from Whisperfish into here.

PR #112, opened 2021-09-15
Generated fields are now (De)Serializable

EcoloSweet:
Sometimes, it might be useful to be able to Serialize and Unserialize fields.

For instance, if you are fetching a message attachment (with an `AttachmentPointer`) and the internet becomes unavailable, it would be great to keep the message in a separate database to be able to get it later.

Doesn't cost much but might be useful for other future uses.

### DIFF
--- a/libsignal-service/build.rs
+++ b/libsignal-service/build.rs
@@ -28,5 +28,7 @@ fn main() {
         })
         .collect();
 
-    prost_build::compile_protos(&input, &[protobuf]).unwrap();
+    let mut config = prost_build::Config::new();
+    config.type_attribute(".", "#[derive(serde::Serialize, serde::Deserialize)]");
+    config.compile_protos(&input, &[protobuf]).unwrap();
 }


EcoloSweet:
The goal is to serialize the object itself, not the pointer.


Concerning the specific above example:

Imagine you received a message and started downloading attachments you find in the `Vec<AttachmentPointer>`. Then, the internet goes down and the download fails. My idea was to serialize the AttachmentPointer and keep it for later so we don't lost the attachment.

I don't know how long the attachment is kept nor how many times we can get it but I guess that it is possible to retry if the attachment is not fully downloaded.

rubdos:
I'm talking about the object too! It's just that it's called "pointer". Signal changes their `struct AttachmentPointer` inbetween releases. You can Serialize to JSON, and the day after, you can find out that the object doesn't Deserialize anymore, because of a spec change. The protobuf is always kept backwards compatible, so you're probably better off serializing via protobuf (i.e., `Message::encode_to_vec()`), and store that as a binary in your database. Does this make sense?

Specifically to attachments, I think Signal-Android implements a queueing system, both for uploading and for downloading.

EcoloSweet:
Oh ok I see. This is perfect for my need, thank you very much for these information.  I guess I can cancel this merge request :)

rubdos:
Glad we agree! Using the protobuf will help you maintain forward compatibility. By the way, feel free to join #whisperfish:rubdos.be on Matrix if you want to chat about this project!

PR #113, opened 2021-09-30
Add hyper based registration example

jonnius:
Registration needs to be interactive, at least to enter the confirmation code after sending the registration request. Therefore I did not allow for command line arguments, but rather provided mock functions for user interaction.

### DIFF
--- a/libsignal-service-hyper/examples/registering.rs
+++ b/libsignal-service-hyper/examples/registering.rs
@@ -1,0 +1,138 @@
+use std::str::FromStr;
+
+use libsignal_service::configuration::{ServiceCredentials, SignalServers};
+use libsignal_service::prelude::phonenumber::PhoneNumber;
+use libsignal_service::provisioning::{
+    generate_registration_id, ConfirmCodeMessage, ConfirmCodeResponse,
+    ProvisioningManager, VerificationCodeResponse,
+};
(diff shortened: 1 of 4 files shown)


libsignal-service-hyper/examples/registering.rs
```diff
@@ -0,0 +1,138 @@
...
+use libsignal_service_hyper::prelude::HyperPushService;
+
+use rand::RngCore;
+
+#[tokio::main]
+async fn main() {
+    let phonenumber = let_user_enter_phone_number();
+    let password = let_user_enter_password();
+    let use_voice = does_user_want_voice_confirmation();
+    let captcha = let_user_solve_captcha();
```
rubdos, line 21: Since it's an example, you could also opt to ask these things on CLI, then a developer can actually run the example too. If you don't feel like implementing that, that's fine by me!

rubdos (approved):

libsignal-service-hyper/examples/registering.rs
rubdos: Honestly, I'll just merge this as-is. We can doodle on that later.

jonnius: Agreed. If I find time, I might get back to it and provide a PR...

PR #115, opened 2021-11-06
Adjust to signal-protocol removal of HKDF "versions".

boxdot:
signal-protocol removed HKDF "versions" in commit
https://github.com/signalapp/libsignal-client/commit/64ad39c54d1fef54e90586d50af0574fb2eb16fb. Instead, we use the hkdf
crate as suggested there.

### DIFF
--- a/libsignal-service/Cargo.toml
+++ b/libsignal-service/Cargo.toml
@@ -40,7 +40,7 @@ prost-build = "0.8"
 
 [dev-dependencies]
 anyhow = "1.0"
-tokio = { version = "1.0", features = [ "macros", "rt-multi-thread" ] }
+tokio = { version = "1.0", features = [ "macros" ] }
 
 [features]
 prefer-e164 = []
--- a/libsignal-service/Cargo.toml
+++ b/libsignal-service/Cargo.toml
@@ -33,13 +33,14 @@ rand = "0.7"
 
 uuid = { version = "0.8", features = [ "serde" ] }
 phonenumber = "0.3"
+hkdf = "0.11.0"
 
 [build-dependencies]
 prost-build = "0.8"
 
 [dev-dependencies]
 anyhow = "1.0"
-tokio = { version = "1.0", features = [ "macros" ] }
+tokio = { version = "1.0", features = [ "macros", "rt-multi-thread" ] }
 
 [features]
 prefer-e164 = []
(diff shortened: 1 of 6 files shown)


libsignal-service/Cargo.toml
```diff
@@ -33,13 +33,14 @@ rand = "0.7"
...
 phonenumber = "0.3"
+hkdf = "0.11.0"
 
 [build-dependencies]
 prost-build = "0.8"
 
 [dev-dependencies]
 anyhow = "1.0"
-tokio = { version = "1.0", features = [ "macros" ] }
+tokio = { version = "1.0", features = [ "macros", "rt-multi-thread" ] }
```
rubdos, line 43: Why is this needed?

libsignal-service/src/groups_v2/utils.rs
```diff
@@ -9,10 +9,10 @@ pub fn derive_v2_migration_master_key(
...
-    let bytes =
-        hkdf.derive_secrets(group_id, b"GV2 Migration", GROUP_MASTER_KEY_LEN)?;
-    let mut bytes_stack = [0u8; GROUP_MASTER_KEY_LEN];
-    bytes_stack.copy_from_slice(&bytes);
-    Ok(GroupMasterKey::new(bytes_stack))
+
+    let mut bytes = [0; GROUP_MASTER_KEY_LEN];
+    hkdf::Hkdf::<sha2::Sha256>::new(None, group_id)
+        .expand(b"GV2 Migration", &mut bytes)
+        .expect("valid output length"); // TODO
```
rubdos, line 16: What's TODO here? Just documenting the validity of the panic is enough IMO.

rubdos:
Not sure whether the change to AWC is really needed in this MR. Should we move that in another discussion?

libsignal-service/Cargo.toml
gferon: will move the `actix` upgrade in another PR.

rubdos: This rt-multi-thread thing is still in. It doesn't affect the library itself, but it's in dev-deps and I don't see why? :'-)

rubdos: I don't mind this being here, but is it necessary somehow?

rubdos (approved):
LGTM, except that that tokio line in Cargo.toml looks off.

Do we have test coverage (especially test vectors) for some of those HKDF calls? I would guess we do since it's such a low-level primitive, but still.

I'd say squash-on-merge.

gferon:
> I'd say squash-on-merge.

I'll squash on merge, and open a new PR for dependency upgrades.

boxdot:
Thank you both. I got carried away with personal life.

rubdos:
> Thank you both. I got carried away with personal life.

Don't worry a bit. We all do that.

PR #117, opened 2021-12-01
Use new URL scheme for secondary device linking

technillogue:
See: https://github.com/signalapp/Signal-Desktop/commit/8064560a5f649c58a98ee4749851d8c923aecc47#diff-b1c3e328d2990b0ab3c57959f539ceb86e248258450042b593428fde18f50e66


gferon:
Thanks a lot! This also fixes the bug where linked clients show as "Unnamed devices" in the Signal app 🥳

rubdos:
Ah, I should've known this when I implemented the other way around in Whisperfish. Thanks, @technillogue !

PR #123, opened 2022-01-18
Fix device-name when linking as secondary device

### LINKED ISSUE #121: Linked secondary devices show as Unnamed device
It looks like we should encrypt the device name, but I'm not sure yet whether this is exactly the problem or not. (This stopped working correctly a few months ago.)

gferon:
Also ~refactored~emptied a function that wasn't really useful, which will break compilation of downstream clients.

Resolves #121

Workaround before #124

### DIFF
--- a/libsignal-service/src/provisioning/manager.rs
+++ b/libsignal-service/src/provisioning/manager.rs
@@ -46,7 +46,8 @@ pub struct ConfirmDeviceMessage {
     pub supports_sms: bool,
     pub fetches_messages: bool,
     pub registration_id: u32,
-    pub name: String,
+    // FIXME: the name goes back here when we send this via the websocket
+    //pub name: String,
 }
 
 impl ConfirmCodeMessage {
@@ -281,7 +282,6 @@ impl<P: PushService> LinkingManager<P> {
         &mut self,
         csprng: &mut R,
         signaling_key: SignalingKey,
-        device_name: &str,
         mut tx: Sender<SecondaryDeviceProvisioning>,
     ) -> Result<(), ProvisioningError> {
         // open a websocket without authentication, to receive a tsurl://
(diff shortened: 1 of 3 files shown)


libsignal-service/src/provisioning/manager.rs
```diff
@@ -46,7 +46,8 @@ pub struct ConfirmDeviceMessage {
     pub supports_sms: bool,
     pub fetches_messages: bool,
     pub registration_id: u32,
-    pub name: String,
+    // FIXME: the name goes back here when we send this via the websocket
+    //pub name: String,
```
rubdos, line 50: Should this become an issue?

rubdos (approved):
LGTM. The FIXME that you introduces should probably be tracked as an issue, I suppose.

rubdos:
OK, I'll merge it now.

rubdos:
```
error[E0061]: this function takes 3 arguments but 4 arguments were supplied
   --> libsignal-service-actix/examples/link.rs:54:27
    |
54  |         provision_manager.provision_secondary_device(
    |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 3 arguments
55  |             &mut csprng,
    |             -----------
56  |             signaling_key,
    |             -------------
57  |             &args.device_name,
    |             -----------------
58  |             tx,
    |             -- supplied 4 arguments
    |
note: associated function defined here
   --> /home/rsmet/src/libsignal-service-rs/libsignal-service/src/provisioning/manager.rs:281:18
    |
281 |     pub async fn provision_secondary_device<R: rand::Rng + rand::CryptoRng>(
    |
```

Seems like we didn't check the examples :'-)

gferon:
CI did fail though 😅 I'll fix it ASAP

PR #125, opened 2022-01-23
Fix compilation

gferon:
@rubdos what about making CI checks mandatory? 🏁


libsignal-service/src/sender.rs
```diff
@@ -127,7 +127,7 @@ pub enum MessageSenderError {
 impl<Service, S, I, SP, P, R> MessageSender<Service, S, I, SP, P, R>
 where
     Service: PushService + Clone,
-    S: SessionStore + SessionStoreExt + Clone,
+    S: SessionStore + SessionStoreExt + Sync + Clone,
```
gferon, line 130: I'm not _exactly_ sure why the `Sync` requirement becomes mandatory here, though.

PR #126, opened 2022-01-23
Fix registration

gferon:
* On registration, `AccountAttributes` are now directly serialized and sent to /v1/accounts/code/
* Truncate the unidentified access key to 16 bytes (see: https://github.com/signalapp/Signal-Server/commit/e6237480f88ff95c56a34ad4bebd62da5c1a070b)

TODO: 
- [x] Rebase after #125 is reviewed and merged


rubdos:
> On registration, `AccountAttributes` are now directly serialized and sent to /v1/accounts/code/

I feel like this only works for secondary devices, could that be the case?

PR #129, opened 2022-01-29
Log sanitization

### LINKED ISSUE #53: Log censoring
Currently, the trace log contains secrets, such as the phone number and signaling password. We might want to filter these out.

@gferon We could introduce a `feature = "unhygienic-logging"`, or we could do `cfg!(debug-assertions)`, to still allow to log these things. What do you think? I'd vote that a debug build can contain sensitive information, while release mode should really filter and censor these crucial things.

rubdos:
I've introduced a new crate [`desan`](https://gitlab.com/whisperfish/desan) under MIT+Apache. `desan` allows to write `#[derive(SanitizeDebug)]`, which derives `Debug` with additional filters for e.g. phone numbers or UUIDs or generic strings.

TODO:

- [ ] Sync desan to the Whisperfish group on Github for visibility.
- [ ] Implement [sanitization for Optionals](https://gitlab.com/whisperfish/desan/-/merge_requests/1) in desan.
- [ ] Add desan to more places: go over everything that derives Debug.
- [ ] Sanitize log statements: go over every `{:?}`-style format and check whether it's either Desan or clean of its own.

FWIW, desan is a "purely compile time dependency" and is basically just a macro. Shouldn't add any overhead that a manual sanitization implementation would have.

Fixes #53

### DIFF
--- a/libsignal-service-actix/src/websocket.rs
+++ b/libsignal-service-actix/src/websocket.rs
@@ -137,15 +137,25 @@ impl AwcWebSocket {
         url.set_scheme("wss").expect("valid https base url");
 
         if let Some(credentials) = credentials {
+            log::trace!(
+                "Will start websocket at {:?} with credentials {:?}",
+                url,
+                credentials
+            );
+
             url.query_pairs_mut()
                 .append_pair("login", credentials.login().as_ref())
                 .append_pair(
                     "password",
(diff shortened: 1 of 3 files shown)


rubdos:
Closing in favor of tracing.

PR #132, opened 2022-02-04
Add key for staging servers root CA

### LINKED ISSUE #131: Invalid certificate when using SignalServers::Staging
I'm not sure if this is already known/intended, but I couldn't get a device link URL without switching to the Production servers.  I get the following error on Staging:
> Service error: Websocket error: error while connecting to websocket: IO error: invalid certificate: UnknownIssuer

I see you have SIGNAL_ROOT_CA hard-coded in the lib, so I thought something might have changed upstream.  Do you do most of your development against the production servers?  Have you tried to use something like [dosco/signal-dev](https://github.com/dosco/signal-dev)?

Edit: I've attached the certificate I downloaded manually from https://chat.staging.signal.org/ and it is indeed different than the one for https://chat.signal.org/ which is the same as the hard-coded one.

[signal-cert.zip](https://github.com/whisperfish/libsignal-service-rs/files/7998086/signal-cert.zip)

hackel:
Resolves #131

### DIFF
--- a/libsignal-service/src/configuration.rs
+++ b/libsignal-service/src/configuration.rs
@@ -64,8 +64,12 @@ impl ServiceCredentials {
     }
 }
 
-const SIGNAL_ROOT_CA: &str = r#"-----BEGIN CERTIFICATE-----
+const SIGNAL_PROD_ROOT_CA: &str = r#"-----BEGIN CERTIFICATE-----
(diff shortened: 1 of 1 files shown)


hackel:
Oops, this was premature.  It only works for chat.staging.signal.org, not any of the other servers.  I think I put the server cert in there by mistake.  Investigating.

PR #136, opened 2022-02-12
Introduce MSRV

### LINKED ISSUE #135: [CI] Enforce MSRV
For SailfishOS, we require MSRV = 1.52. It would be great if we could enforce this in CI.

rubdos:
Fixes #135


gferon (approved):

rubdos:
@gferon, is the hyper build failure on 1.52 something we want to fix? We could also just enforce 1.52 on Actix...

PR #139, opened 2022-03-11
SessionStoreExt::compute_safety_number

rubdos:
Helper method to bundle all the necessary SessionStore requests into computation of the safety number. Maybe some day we should get rid of the magical constants in here. 5200, really?

### DIFF
--- a/libsignal-service/src/service_address.rs
+++ b/libsignal-service/src/service_address.rs
@@ -96,6 +96,18 @@ impl ServiceAddress {
         (self.phonenumber.is_some() && self.phonenumber == other.phonenumber)
             || (self.uuid.is_some() && self.uuid == other.uuid)
     }
+
+    /// Returns e164 if present, uuid otherwise.
+    pub fn e164_or_uuid(&self) -> String {
+        if let Some(e164) = self.e164() {
+            return e164;
+        } else if let Some(ref uuid) = self.uuid {
+            return uuid.to_string();
+        }
+        unreachable!(
+            "an address requires either a UUID or a E164 phone number"
+        );
+    }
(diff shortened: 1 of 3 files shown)


rubdos:
@gferon, when you'd review, have a look at the new clippy warning about SessionExt not being able to be made into an object.

gferon:
> SessionExt not being able to be made into an object.

At this stage, since we use those functionalities everywhere, what about simply merging the traits?

libsignal-service/src/service_address.rs
```diff
@@ -96,6 +96,18 @@ impl ServiceAddress {
...
+
+    /// Returns e164 if present, uuid otherwise.
+    pub fn e164_or_uuid(&self) -> String {
+        if let Some(e164) = self.e164() {
+            return e164;
+        } else if let Some(ref uuid) = self.uuid {
+            return uuid.to_string();
+        }
+        unreachable!(
+            "an address requires either a UUID or a E164 phone number"
```
gferon, line 108: technically since `ServiceAddress` is public, you could build it in an invalid state... do we really want this? (I know we have a few panics here and there in the lib, but maybe we can try avoiding new ones?).

gferon (approved):
LGTM, one minor comment that you can disregard if you think that's ok.

rubdos:
> 

The reason we have an extension trait is because the other traits are in libsignal-client, no?

libsignal-service/src/service_address.rs
rubdos: Hmm, then I'd rather have the initializer set to private at some point in the future.

PR #140, opened 2022-03-21
Use static dispatch for SessionStoreExt in sub_device_sessions

gferon:
This fixes the warning explained in https://github.com/rust-lang/rust/issues/51443.


rubdos (approved):
LGTM

PR #142, opened 2022-04-09
[CI] Enforce toolchain version

rubdos:


rubdos:
[57c7c16](https://github.com/whisperfish/libsignal-service-rs/pull/142/commits/57c7c16bbbee79c3d71b06af2676097c67c6b154) is a hack and I don't want it. It does make everything build again on 1.52, but it also forces lower quote/async-trait/proc-macro2 versions for people with more recent compilers. Suggestion, @gferon ?

rubdos:
This also adds building on nightly, because otherwise the matrix was so empty :|

rubdos:
New version of the hack ([a1a8894](https://github.com/whisperfish/libsignal-service-rs/pull/142/commits/a1a889485e126a694082276c00701e70d843c7d9)) is more hacky, but also less intrusive. It does however, again, violate additive features. In the end, we only would use this in the CI however.

gferon:
I think it's perfectly fine (and quite smart, I must say) like this for now, we can always remove this small hack when we bump the MSRV.

Suggestion: any particular reason the nightly version is hardcoded?

gferon (approved):

rubdos:
> Suggestion: any particular reason the nightly version is hardcoded?

Because apparently the Rust version that SailfishOS uses is 1.52.1, but built as a nightly, not as a stable...

PR #143, opened 2022-04-09
Update libsignal-client to v0.16

rubdos:
There were quite a few changes to sealed session that we need to incorporate. Let's make an issue about that, or fix them here.

* Remove our own implementation of sealed sender messages and use the one from `libsignal-client`
* Bump `libsignal-client` to 0.16.0


gferon:
> * There were quite a few changes to sealed session that we need to incorporate. Let's make an issue about that, or fix them here.

Most of the code I added in there is duplicated in `libsignal-client`, right? I could try to add commits here to remove it.

rubdos:
> Most of the code I added in there is duplicated in `libsignal-client`, right? I could try to add commits here to remove it.

I'm not sure, but you know this part of the code base a lot better than I do. Feel free to work on top of this! Maybe it's already useful to merge this before, if it's backwards compatible (if I figure out why the build system failed). You know, merge early and break things, or something like that.

gferon:
@rubdos you can't review your own PR, but I completely removed `sealed_session_cipher` _and_ "implemented" sealed message encryption 🥳

gferon:
Forgot to mention that we should bump the version of the crate because there's a minor breaking API change.

rubdos:
> Forgot to mention that we should bump the version of the crate because there's a minor breaking API change.

Have we ever done that?

gferon:
> > Forgot to mention that we should bump the version of the crate because there's a minor breaking API change.
> 
> Have we ever done that?

Haha, no, but might as well start at some point - I feel we don't really break the API as often anymore.

PR #144, opened 2022-05-03
Update groups v2 support

gferon:
* Update the Protobuf schemas of groups v2 to the latest available version and make relevant code changes, which will add support for all (current 📅) features of groups v2.
* Add the basic bricks (essentially, a [structured enum for decrypted `GroupChange`](https://github.com/whisperfish/libsignal-service-rs/pull/144/files#diff-3f6d16398a076979cc60d7642747d3dac2417f204fd13a7235bb66c717fad0e4R79) to build a state machine and properly handle group changes
* Fix some unused fields warnings (completely unrelated to this PR, but my 👀 were bleeding a little bit when compiling)
* Bump `libsignal-client` to `v0.17.0` (the latest release as of this PR)

TODO: 
- [x] Test it 😆


libsignal-service/src/push_service.rs
```diff
@@ -255,25 +255,9 @@ pub struct SignalServiceProfile {
...
-    key: String,
-    policy: String,
-    algorithm: String,
-    credential: String,
-    date: String,
-    signature: String,
-    content_type: String,
-    length: u64,
-}
-
```
rubdos, line 272: Is this unused now?

libsignal-service/src/groups_v2/operations.rs
```diff
@@ -44,99 +230,95 @@ impl From<zkgroup::ZkGroupVerificationFailure> for GroupDecryptionError {
...
+            *decrypted_uuid.as_bytes(),
         )?)
     }
 
+    fn decrypt_profile_key_presentation(
+        &self,
+        presentation: &[u8],
+    ) -> Result<(Uuid, ProfileKey), GroupDecryptionError> {
+        let profile_key_credential_presentation =
+            AnyProfileKeyCredentialPresentation::new(&presentation)?;
```
rubdos, line 256:
```
 this expression creates a reference which is immediately dereferenced by the compiler

warning: this expression creates a reference which is immediately dereferenced by the compiler
   --> libsignal-service/src/groups_v2/operations.rs:256:54
    |
256 |             AnyProfileKeyCredentialPresentation::new(&presentation)?;
    |                                                      ^^^^^^^^^^^^^ help: change this to: `presentation`
    |
    = note: `#[warn(clippy::needless_borrow)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

```

rubdos (approved):
LGTM. One question in the review, one here:

What you implemented is the **decryption** of the group delta's, right? Not really the application of the change onto a group, together with it's validation? The state machine itself should still follow, from what I understand here.

libsignal-service/src/push_service.rs
gferon: Seems to be, yes. I cleaned up compilation warnings in the PR.

gferon:
> What you implemented is the **decryption** of the group delta's, right? Not really the application of the change onto a group, together with it's validation? The state machine itself should still follow, from what I understand here.

Exactly, however the new function collects the group changes in a nice Rust-y enum that makes it super easy to `match` on. I don't think we can do the state machine inside of `libsignal-service` without introducing a trait for storing group information (I'm open to the idea, but this is a bigger undertaking than going forward with the protocol upgrades).

PR #145, opened 2022-05-11
Bump libsignal-client to v0.14.0

gferon:
Bump to the maximum possible version that doesn't introduce breaking changes, before starting updating and refactoring groups v2.


rubdos:
This still bounces on `array_map` https://github.com/whisperfish/libsignal-service-rs/runs/6383535286?check_suite_focus=true , but it's something I'm willing to patch out in a forked version for Whisperfish.

rubdos (approved):

PR #148, opened 2022-06-23
Use &mut self for write operations of SessionStoreExt trait

gferon:
I implemented some in-memory storage and came across this small mistake I did when I designed the trait. Of course, one can use interior mutability, but it's a bit of a shame 😄


rubdos:
Clippy did a better review than I could :thinking: 

Is that a regression due to your commit? If not, I guess we merge this.

rubdos:
> Yes, clippy just reminded me of the whole "unsend" story - I completely forgot about it. Closing this.

Well, I'm not sure that that's right; I suppose this needs some re-engineering work anyway.

PR #150, opened 2022-07-12
Fix sealed sender decryption

gferon:
I mixed up the local UUID and device_id with the recipient's one... I really don't know what went through my mind at the time 😓 

You can see it now works because the message correctly deserializes (it didn't before, and actually never even reached this code).

```
[2022-07-12T14:40:10Z DEBUG libsignal_protocol::sealed_sender] deserializing UnidentifiedSenderMessage with version 1
[2022-07-12T14:40:10Z INFO  libsignal_protocol::sealed_sender] deserialized UnidentifiedSenderMessageContent from 779a6c6b-0985-479a-86fc-ce435a3db130.20 with type PreKey
[2022-07-12T14:40:10Z DEBUG libsignal_protocol::session_cipher] 779a6c6b-0985-479a-86fc-ce435a3db130.20 has existing receiver chain.
[2022-07-12T14:40:10Z INFO  libsignal_protocol::session_cipher] decrypted PreKey message from 779a6c6b-0985-479a-86fc-ce435a3db130.20 with current session state (base key e017108322351554646c10d04f011c4b67bb83cb52acfed23f4429e005f9c90f)
```

This will fix issues like https://github.com/whisperfish/presage/issues/62

### DIFF
--- a/libsignal-service/src/cipher.rs
+++ b/libsignal-service/src/cipher.rs
@@ -144,7 +144,10 @@ where
                     .await?
                     .ok_or(SignalProtocolError::SessionNotFound(sender))?;
 
-                strip_padding_version(session_record.session_version()?, &mut data)?;
+                strip_padding_version(
+                    session_record.session_version()?,
+                    &mut data,
+                )?;
                 Plaintext { metadata, data }
             },
             Type::Ciphertext => {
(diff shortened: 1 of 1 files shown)


rubdos:
It shows that Whisperfish does not yet use SS... Shame on me.

gferon:
@rubdos figured out decrypted sealed sender messages are padded. Progress has been made, this should be good to go!

libsignal-service/src/cipher.rs
```diff
@@ -213,15 +215,19 @@ where
...
                     timestamp: envelope.timestamp(),
                     needs_receipt: false,
                 };
+
+                strip_padding(&mut message)?;
+
                 Plaintext {
-                    metadata,
-                    data: message,
+                    metadata: metadata,
```
rubdos, line 229: clippy.

gferon:
> This looks simple enough. Seems like you also tested it, so LGTM :-)

Might not be good enough since I "tested" the previous one as well 😂 can't break what's broken though.

PR #119, opened 2022-01-09
Implement device name encryption and decryption

gferon:
This was introduced in Signal-Desktop a few months ago. I'll implement encryption and decryption. In the meantime to make sure people can properly register their devices, we should stop passing a value to the `name` field.

TODO:

- [ ] Use websocket to call the provisioning API

Resolves #124

PR #151, opened 2022-08-17
Fix device linking

Schmiddiii:
The reason behind <https://github.com/whisperfish/libsignal-service-rs/pull/119#issuecomment-1217704738> was a rouge second `Self::json`. This would lead to the whole response being consumed by the first `Self::json` to print the response, the second `Self::json` would then not have to decode a empty response.


gferon:
Thanks a lot for the investigation and the fix!

PR #152, opened 2022-08-19
Update AccountAttributes/DeviceCapabilities from Signal Android

rubdos:
Some fields were removed in 4876e9555, while still present in Android. @gferon, could you elaborate why you removed them?

Additionally, I would want to point out that the same commit changed an example, in which `DeviceCapabilities::default()` is now used instead of manually setting capabilities. Maybe it should document that this is not how a real application should behave.

### DIFF
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -106,10 +106,13 @@ pub struct DeviceCapabilities {
     pub announcement_group: bool,
     #[serde(rename = "gv2-3")]
     pub gv2: bool,
+    pub storage: bool,
     #[serde(rename = "gv1-migration")]
     pub gv1_migration: bool,
     pub sender_key: bool,
     pub change_number: bool,
+    pub stories: bool,
+    pub gift_badges: bool,
 }
 
 #[derive(Clone)]


gferon:
I'm not super sure why I did this anymore, but there's some discrepancy between the way we confirm registration here and in the Java implementation IIRC - we use the synchronous HTTP API and they use the websocket connection.

tbh, I can't really recall what the exact differences are, so if this works we're good to go. Thanks for figuring it out!

gferon:
> I assume then you are looking at Signal Desktop

Indeed. At least for this specific change I looked at `Signal-Desktop`.

rubdos:
Aha, looking at Signal-Server now.

rubdos:
I cannot seem to find a trace of these fields any more. I assume that the Websocket connection is friendlier and strips out unknown fields. I'll alter the MR to reflect the current state of the server. This is quite an annoying situation though, because it'll break any time the server changes without any transition period. We'll need to move the registration into the websocket. Let's make an issue about that before we merge this.

rubdos:
> Could it be that the sync endpoint actually does the same as the websocket one behind the scenes now (as opposed to a few months ago)?

That might very well be the case, but I'm not really in the mood to go dig and find it out.

rubdos:
Okay I don't find any evidence in the server code that the `storage` device capability has been removed.  I also don't find any evidence that Signal-Desktop ever had this field.

How do you go about testing this; just try and register? Via the registration example, maybe?

rubdos:
I've dropped the PNI field for now, since I don't see a sane default for it if we don't implement phone number identifiers also.

rubdos:
@gferon, this is still untested, but since it's been cross-checked with Signal-Server and Signal-Android, I'm inclined to merge. What do you think?

gferon:
Sorry, I thought this was merged already 🙈 go ahead!

PR #154, opened 2022-09-20
Upgrade libsignal-protocol 0.17.0 -> 0.20.0

boxdot:


libsignal-service/src/push_service.rs
```diff
@@ -221,14 +221,17 @@ impl PreKeyResponseItem {
     ) -> Result<PreKeyBundle, SignalProtocolError> {
         PreKeyBundle::new(
             self.registration_id,
-            self.device_id,
+            self.device_id.into(),
```
boxdot, line 224: I could propage the strong types on the service level. But this would be breaking change then. WDYT?

rubdos: I don't think we really shy away from breaking the interface, especially since it broke in libsignal-client already. Would you be okay with me merging #156 first, then rebase this on top?

libsignal-service/Cargo.toml
```diff
@@ -12,14 +12,13 @@ license = "GPLv3"
 readme = "../README.md"
 
 [dependencies]
-libsignal-protocol = { git = "https://github.com/signalapp/libsignal-client", tag = "v0.17.0" }
-zkgroup = { git = "https://github.com/signalapp/libsignal-client", tag = "v0.17.0" }
+libsignal-protocol = { git = "https://github.com/signalapp/libsignal-client", tag = "v0.20.0" }
+zkgroup = { git = "https://github.com/signalapp/libsignal-client", tag = "v0.20.0" }
```
iamdavidcz, line 16: @boxdot
Apparently, `https://github.com/signalapp/libsignal-client` now redirects to `https://github.com/signalapp/libsignal`.
Could you update this too, please?

PR #155, opened 2022-09-20
Throw a compilation error when hyper is compiled with unsend-futures

Schmiddiii:
This throws a descriptive compilation error when libsignal-service-hyper is compiled with unsend-futures like discussed in the Matrix.


rubdos:
@gferon, doesn't #160 include a fix such that libsignal-service-hyper can be compiled *with* this feature?

rubdos:
Closing in favour of the little patch in #160

PR #156, opened 2022-09-21
Don't send messages to recipients default device twice

direc85:
I'm not sure if this is the correct place to correct the issue, or if the logic behind this is even correct, but it seems to work for me.

For some time, Whisperfish seems to be sending double messages to some recipients. I believe this is what happens:

- Send the message to the default device ID (in my Whisperfish, this is always `1`)
- Get the list of sub-device session IDs (list of all IDs: `1,2,3...`)
- Send the message to all IDs
- Default device ID gets the message sent twice

I tested this with both p2p conversations (with myself using my spare SIM, and another contact) and group messages, and it seems to prevent duplicates - but only for the most part. One user of Flare reported still getting duplicates, so I don't really know what could be causing that.

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -632,6 +632,7 @@ where
         content: &[u8],
     ) -> Result<Vec<OutgoingPushMessage>, MessageSenderError> {
         let mut messages = vec![];
+        let mut skip_default_device = false;
 
         let myself = recipient.matches(&self.local_address);
         if !myself || unidentified_access.is_some() {
@@ -645,11 +646,15 @@ where
                 )
                 .await?,
             );
+            skip_default_device = true;
         }
 
         for device_id in
             recipient.sub_device_sessions(&self.session_store).await?
         {
+            if skip_default_device && device_id == DEFAULT_DEVICE_ID {
+                continue
+            }
             trace!("sending message to device {}", device_id);
             let ppa = get_preferred_protocol_address(
                 &self.session_store,
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -653,7 +653,12 @@ where
         devices
             .extend(recipient.sub_device_sessions(&self.session_store).await?);
         devices.sort_unstable();
+
+        let original_device_count = devices.len();
         devices.dedup();
+        if devices.len() != original_device_count {
+            log::warn!("SessionStoreExt::get_sub_device_sessions should return unique device id's, and should not include DEFAULT_DEVICE_ID.");
+        }
 
         // When sending to ourselves, don't include the local device
         if myself {
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -632,6 +632,7 @@ where
         content: &[u8],
     ) -> Result<Vec<OutgoingPushMessage>, MessageSenderError> {
         let mut messages = vec![];
+        let mut devices = vec![];
 
         let myself = recipient.matches(&self.local_address);
         if !myself || unidentified_access.is_some() {
@@ -645,11 +646,13 @@ where
                 )
                 .await?,
             );
+        } else {
+            devices.push(DEFAULT_DEVICE_ID);
         }
 
-        let mut devices = vec![DEFAULT_DEVICE_ID];
         devices
             .extend(recipient.sub_device_sessions(&self.session_store).await?);
+        devices.dedup();
 
         // When sending to ourselves, don't include the local device
         if myself {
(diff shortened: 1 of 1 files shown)


rubdos:
What about this? This is what happens in Android:

```diff+rust
diff --git a/libsignal-service/src/sender.rs b/libsignal-service/src/sender.rs
index b25b3924..b9d143a4 100644
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -632,7 +632,6 @@ where
         content: &[u8],
     ) -> Result<Vec<OutgoingPushMessage>, MessageSenderError> {
         let mut messages = vec![];
-        let mut skip_default_device = false;

         let myself = recipient.matches(&self.local_address);
         if !myself || unidentified_access.is_some() {
@@ -646,15 +645,25 @@ where
                 )
                 .await?,
             );
-            skip_default_device = true;
         }

-        for device_id in
-            recipient.sub_device_sessions(&self.session_store).await?
-        {
-            if skip_default_device && device_id == DEFAULT_DEVICE_ID {
-                continue
-            }
+        let mut devices = vec![DEFAULT_DEVICE_ID];
+        devices
+            .extend(recipient.sub_device_sessions(&self.session_store).await?);
+
+        // When sending to ourselves, don't include the local device
+        if myself {
+            devices.retain(|id| *id != self.device_id);
+        }
+
+        devices.sort_unstable();
+        let original_device_count = devices.len();
+        devices.dedup();
+        if devices.len() != original_device_count {
+            log::warn!("SessionStoreExt::get_sub_device_sessions should return unique device id's, and should not include DEFAULT_DEVICE_ID.");
+        }
+
+        for device_id in devices {
             trace!("sending message to device {}", device_id);
             let ppa = get_preferred_protocol_address(
                 &self.session_store,
```

boxdot:
Isn't this what I basically did in #108, except this is done higher up in the call tree?

direc85:
> Isn't this what I basically did in #108, except this MR does it higher up in the call tree?

The issue is that the message is _may be sent_ to the DEFAULT_DEVICE_ID, and then the DEFAULT_DEVICE_ID is included again in `sub_device_sessions()` results. I'm working on a correct-er fix.

direc85:
This works now in conjunction with [Exclude DEFAULT_DEVICE_ID from sub device session results](https://gitlab.com/whisperfish/whisperfish/-/merge_requests/298) to prevent sending double messages.

libsignal-service/src/sender.rs
```diff
@@ -645,23 +646,20 @@ where
...
                 .await?,
             );
+        } else {
+            devices.push(DEFAULT_DEVICE_ID);
         }
 
-        let mut devices = vec![DEFAULT_DEVICE_ID];
         devices
             .extend(recipient.sub_device_sessions(&self.session_store).await?);
+        devices.dedup();
```
rubdos, line 655: dedup only works for consecutive duplicates, and should therefor come after the `sort_unstable`. Is there a reason you altered that?

libsignal-service/src/sender.rs
```diff
@@ -645,23 +646,20 @@ where
...
         if myself {
             devices.retain(|id| *id != self.device_id);
         }
 
         devices.sort_unstable();
-        let original_device_count = devices.len();
-        devices.dedup();
-        if devices.len() != original_device_count {
-            log::warn!("SessionStoreExt::get_sub_device_sessions should return unique device id's, and should not include DEFAULT_DEVICE_ID.");
-        }
```
rubdos: I thought this would be a useful warning/runtime check. What's the reason behind removing it?

rubdos (commented):
One minor thing, one question.

libsignal-service/src/sender.rs
direc85: I wasn't aware that `dedup()` works that way; I'll change it back the way it was.

libsignal-service/src/sender.rs
direc85: I think it got lost somewhere along the way... It's a useful check to have, I'll put it back.

rubdos:
I think this needs another real-world test and then we'd be good to go, right? :-)

rubdos:
> Isn't this what I basically did in #108, except this MR does it higher up in the call tree?

The premise was different, back then. You had overlapping sessions, and that shouldn't happen. This is not due to overlapping sessions, but due to the application supplying the exact same session twice. You're right that it's the same logic, however.

I'll merge this now, @Schmiddiii just tested it.

rubdos:
Cool, CI fails because of Rust 2021. Someone thought that was a minor version bump. Let's fix that somewhere else.

PR #160, opened 2022-09-25
Fix rustdoc

gferon:


Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
     "libsignal-service-actix",
     "libsignal-service-hyper",
 ]
+default-members = ["libsignal-service", "libsignal-service-hyper"]
```
gferon, line 7: I'm sorry, I had to make a choice (I thread a 🪙)

rubdos: Hmm, if you use `unsend-futures`, can't you have all three?

gferon: It's not really used actually, this is just a 💩 trick to have `cargo check` do... something.

PR #161, opened 2022-09-25
Disable oldtime feature of chrono

gferon:
This will make us benefit from the fix to https://rustsec.org/advisories/RUSTSEC-2020-0071

### DIFF
--- a/libsignal-service/Cargo.toml
+++ b/libsignal-service/Cargo.toml
@@ -23,7 +23,7 @@ thiserror = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 prost = "0.10"
 http = "0.2"
-chrono = { version = "0.4", features = ["serde"] }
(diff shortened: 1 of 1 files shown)


rubdos:
> This will make us benefit from the fix to https://rustsec.org/advisories/RUSTSEC-2020-0071

Do you have a link on how this related to chrono? It's strange that you need to change the feature set to patch out a vulnerability.

gferon:
I think the whole ecosystem decided to use the newer `time` crate or not, but the change is a breaking one, which is why some projects decided to have the fix opt-in.

rubdos:
> I think the whole ecosystem decided to use the newer `time` crate or not, but the change is a breaking one, which is why some projects decided to have the fix opt-in.

Okay, so we can maybe decide to break later (when Diesel works with `time`) and merge this as-is.

PR #163, opened 2022-10-03
Profile fetching update

rubdos:
Depends on #170, although it shouldn't *really*.

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -14,6 +14,7 @@ use serde::{Deserialize, Serialize};
 use sha2::Sha256;
 use zkgroup::profiles::ProfileKey;
 
+use crate::ServiceAddress;
 use crate::{
     configuration::{Endpoint, ServiceCredentials},
     pre_keys::{PreKeyEntity, PreKeyState},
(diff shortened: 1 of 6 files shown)


libsignal-service/src/sender.rs
```diff
@@ -663,7 +663,7 @@ where
 
         // When sending to ourselves, don't include the local device
         if myself {
-            devices.retain(|id| *id != self.device_id.into());
+            devices.retain(|id| DeviceId::from(*id) != self.device_id);
```
rubdos, line 666: For some reason, this comparison started complaining when I implemented the `SignalWebSocket::retrieve_profile_by_id` method. Commenting it out, it worked, but including it triggered some shenanigans around the `.into()` call needing disambiguation.

libsignal-service/src/account_manager.rs
```diff
@@ -313,38 +314,19 @@ impl<Service: PushService> AccountManager<Service> {
...
-            .map(|data| profile_cipher.decrypt_emoji(data))
-            .transpose()?;
-
-        Ok(Profile {
-            name,
-            about,
-            about_emoji,
-        })
+        let profile_cipher = ProfileCipher::from(profile_key);
+        Ok(encrypted_profile.decrypt(profile_cipher)?)
```
rubdos, line 329: I've moved this out of the AccountManager into an arguably *worse* spot. The `AccountManager` should most probably be rewritten against the `SignalWebSocket` now, but that probably implies we want some unified interface between `PushService` and `SignalWebSocket`. In principle they should support the same calls...

rubdos:
Somehow Github managed to merge this without noticing it's merged. Strange stuff.

PR #164, opened 2022-10-04
Fix compilation

gferon:
Fix expected `u32`, found struct `libsignal_protocol::DeviceId`


rubdos:
... sorry about that.

PR #165, opened 2022-10-07
Bump uuid to ~1

rubdos:
libsignal-protocol is on this version now, gives version clashes.


rubdos:
Move fast and break stuff. Merging.

PR #169, opened 2022-10-31
Make Group serde

gferon:
* Make Group struct Serialize and Deserialize
* Expose types necessary to implement CredentialsStore


rubdos:
I can feel sender keys in this request :thinking:

rubdos (changes requested):
CI fails, but apart from that this looks fine.

PR #170, opened 2022-11-04
Refactor common websocket code into a SignalWebSocket system

rubdos:
TODO:
- [x] Reimplement keep-alive in `SignalWebSocket`.
- ~~[ ] Move keep-alive trigger code over to SignalWebSocket, if possible. I think it is possible...~~ Never mind, time is executor-dependent.
- ~~[ ] Clean-up pass over MessagePipe and ProvisioningPipe~~ I don't think we can make these more minimal.

This lays the foundation of sending messages via the WebSocket instead of a separate REST endpoint, and thus supersedes #101.

### DIFF
--- a/libsignal-service/src/websocket.rs
+++ b/libsignal-service/src/websocket.rs
@@ -1,0 +1,330 @@
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex, MutexGuard};
+
+use std::future::Future;
+
+use bytes::Bytes;
+use futures::channel::oneshot::Canceled;
+use futures::channel::{mpsc, oneshot};
+use futures::future::BoxFuture;
+use futures::stream::FuturesUnordered;
+use futures::{SinkExt, Stream, StreamExt};
+use prost::Message;
+
+use crate::messagepipe::{WebSocketService, WebSocketStreamItem};
+use crate::proto::{
+    web_socket_message, WebSocketMessage, WebSocketRequestMessage,
+    WebSocketResponseMessage,
+};
+use crate::push_service::ServiceError;
+
+type RequestStreamItem = (
+    WebSocketRequestMessage,
+    oneshot::Sender<WebSocketResponseMessage>,
+);
+
+pub struct SignalRequestStream {
+    inner: mpsc::Receiver<RequestStreamItem>,
+}
+
+impl Stream for SignalRequestStream {
+    type Item = RequestStreamItem;
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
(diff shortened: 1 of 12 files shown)


rubdos:
One annoyance I have with this design, is that the `SignalWebSocket` could be bound to any remote `wss://` endpoint, and that we cannot guarantee that e.g. `SignalWebSocket::retrieve_profile_by_id` is actually a valid method on a connected `SignalWebSocket`.

We'd also have to think about unidentified websockets vs identified websockets: do we want the application to have to manage both instances? I would think not. Maybe `MessagePipe` should contain both, and should return a stream upon request, instead of implement a stream. This would (and should) break API.

libsignal-service/src/websocket.rs
```diff
@@ -151,6 +151,17 @@ impl<WS: WebSocketService> SignalWebSocketProcess<WS> {
         }
     }
 
+    fn next_request_id(&self) -> u64 {
+        use rand::Rng;
+        let mut rng = rand::thread_rng();
+        loop {
+            let id = rng.gen();
```
SfietKonstantin, line 158: A matter of taste, but for expressivity, you can express that as

"please continue generating an id up until you find one that is not part of outgoing request"

so

```
iter::repeat_with(|| rng.gen()).filter(|id| !self.outgoing_request_map.contains_key(&id)).next()
```

Problem is that it returns an optional, but we might be able to unwrap here

libsignal-service/src/websocket.rs
```diff
@@ -70,6 +70,8 @@ struct SignalWebSocketProcess<WS: WebSocketService> {
         oneshot::Sender<Result<WebSocketResponseMessage, ServiceError>>,
     >,
 
+    outgoing_keep_alive_map: HashSet<u64>,
```
SfietKonstantin, line 74: it's a set !

libsignal-service/src/websocket.rs
```diff
@@ -176,19 +190,28 @@ impl<WS: WebSocketService> SignalWebSocketProcess<WS> {
...
-                            //         });
-                            //         background_work.push(request.boxed_local());
-                            //     },
-                            //     Err(e) => log::warn!("Could not send keep alive: {}", e),
-                            // }
+                            // XXX: would be nicer if we could drop this request into the request
+                            // queue above.
+                            log::debug!("Sending keep alive upon request");
+                            let request = WebSocketRequestMessage {
+                                id: Some(
```
SfietKonstantin: Maybe the robust id generator could be used here, no ?

libsignal-service/src/websocket.rs
rubdos: I was also thinking on `while self.map.contains_key(&id) { id = rand.gen();}`... I don't like the `loop` either.

PR #171, opened 2022-11-05
Implement reception of SenderKeyDistributionMessage and decrypting of groups with sender keys

gferon:


libsignal-service/Cargo.toml
```diff
@@ -12,8 +12,8 @@ license = "GPLv3"
 readme = "../README.md"
 
 [dependencies]
-libsignal-protocol = { git = "https://github.com/signalapp/libsignal", tag = "v0.20.0" }
-zkgroup = { git = "https://github.com/signalapp/libsignal", tag = "v0.20.0" }
+libsignal-protocol = { git = "https://github.com/signalapp/libsignal", tag = "v0.21.1" }
+zkgroup = { git = "https://github.com/signalapp/libsignal", tag = "v0.21.1" }
```
rubdos, line 16: I'll refrain from merging this until I got the profile code in, because I'll have to update my libsignal fork for Whisperfish too with this bump.

libsignal-service/src/cipher.rs
```diff
@@ -78,12 +83,27 @@ where
             let plaintext = self.decrypt(&envelope).await?;
             let message =
                 crate::proto::DataMessage::decode(plaintext.data.as_slice())?;
-            Ok(Some(Content::from_body(message, plaintext.metadata)))
+            let content = Content::from_body(message, plaintext.metadata);
+            Ok(Some(content))
```
rubdos, line 87: I don't see the relevance of this?

libsignal-service/src/cipher.rs
```diff
@@ -78,12 +83,27 @@ where
...
         } else if envelope.content.is_some() {
             let plaintext = self.decrypt(&envelope).await?;
             let message =
                 crate::proto::Content::decode(plaintext.data.as_slice())?;
-            Ok(Content::from_proto(message, plaintext.metadata))
+            if let Some(bytes) = message.sender_key_distribution_message {
+                dbg!("PROCESSING SENDER KEYSSSSSSSS -------------------------");
```
rubdos, line 93: Clippy says it too, but this should probably go away. If you want, use `log::debug!()` etc. :-)

libsignal-service/src/cipher.rs
```diff
@@ -121,7 +141,7 @@ where
                 .await?;
                 let metadata = Metadata {
                     sender: envelope.source_address(),
-                    sender_device: envelope.source_device(),
+                    sender_device: envelope.source_device().into(),
```
rubdos, line 144: This is still a `u32`, no?

libsignal-service/src/cipher.rs
```diff
@@ -161,7 +181,7 @@ where
...
```
rubdos, line 184: Same.

libsignal-service/src/cipher.rs
```diff
@@ -422,3 +443,106 @@ pub async fn get_preferred_protocol_address<S: SessionStore>(
...
+
+/// Decrypt a Sealed Sender message `ciphertext` in either the v1 or v2 format, validate its sender
+/// certificate, and then decrypt the inner message payload.
+///
+/// This method calls [`sealed_sender_decrypt_to_usmc`] to extract the sender information, including
+/// the embedded [`SenderCertificate`]. The sender certificate (signed by the [`ServerCertificate`])
+/// is then validated against the `trust_root` baked into the client to ensure that the sender's
+/// identity was not forged.
+#[allow(clippy::too_many_arguments)]
+async fn sealed_sender_decrypt(
```
rubdos, line 455: I don't immediately see how this works. Should the application call this method when the `ContentBody` matches some specific variant? This function is, as far as I can see, never used (but clippy doesn't call you out on it, so I'm potentially blind).

Since you tested in Presage, I assume this works, for both receiving and sending.

rubdos (changes requested):
Some strange things that clippy picks out and a question. Thanks for the heavy lifting!!

gferon:
this PR is missing using sender keys to send group messages, I'll implement this here as well so that we can _really_ test it.

libsignal-service/src/cipher.rs
gferon: the joys of `printf` debugging

rubdos:
> this PR is missing using sender keys to send group messages, I'll implement this here as well so that we can _really_ test it.

That would be great. I've just started to work towards unidentified sending and sender keys in Whisperfish: https://gitlab.com/whisperfish/whisperfish/-/merge_requests/306

It already compiles against this branch, and I've set the unidentified keys and flag :-)

gferon:
> > this PR is missing using sender keys to send group messages, I'll implement this here as well so that we can _really_ test it.
> 
> That would be great. I've just started to work towards unidentified sending and sender keys in Whisperfish: https://gitlab.com/whisperfish/whisperfish/-/merge_requests/306
> 
> It already compiles against this branch, and I've set the unidentified keys and flag :-)

Actually, it makes sense to merge support for receiving as fast as possible, sending will take a bit more time as usual. 

I feel like we need to have clients in the wild that can receive for a few weeks _before_ we switch to sending using sender keys. Wdyt?

libsignal-service/src/cipher.rs
gferon: `Content::from_proto` now returns a `Result` so we can precisely see in the error logs when a new type of message comes in that we don't support (we returned `None` before which was IMO misleading). Let me double check if there's a better way to do this.

libsignal-service/src/cipher.rs
gferon: Indeed, I started to change it to `DeviceId` everywhere and reverted because it's somewhat unrelated. This is just a remnant, I just filed https://github.com/whisperfish/libsignal-service-rs/issues/173.

libsignal-service/src/cipher.rs
gferon: This is (unfortunately) a copy-pasta of the current impl in `libsignal` and is used here https://github.com/whisperfish/libsignal-service-rs/blob/34e24634e2bb66b48f703f619acaed870a3cd475/libsignal-service/src/cipher.rs#L219 the same way we should use the function in `libsignal`, which somehow doesn't have a branch for `CiphertextMessageType::SenderKey` https://github.com/signalapp/libsignal/blob/main/rust/protocol/src/sealed_sender.rs#L1647

gferon:
> Agreed. Then again, I don't have sealed sending yet in WF. I'd like to get that in ASAP too, and then I can push receiving via Sender Keys at the same time, and start working on sending via Sender Keys.

Exactly, it's actually fine to have sealed sender enabled, receive sender keys and groups encrypted with sender keys, without being able to send with them (at least for now, it works). I'm up for doing the sending part as well after that!

libsignal-service/src/cipher.rs
boxdot: I don't quite understand why this is another function. I checked the sources of Signal Desktop. And there this code is part of `open_envelope`.

boxdot: Basically, what they do there: 

1. Unseal the envelope with `sealedSenderDecryptToUsmc`
2. Check the `msgtype` (here is should be part of case for unidentified sender)
3. Depending on the type of `msgtype` call `decrypt` or `groupDecrypt`

boxdot: Ah, now I got it. Basically, you are doing it one level deeper compared to Signal Desktop, which actually makes sense.

rubdos (approved):

gferon:
If/when https://github.com/signalapp/libsignal/pull/494 is merged, we can remove most of the copy-pasted code here with a bump of `libsignal`

PR #172, opened 2022-11-06
impl Display for ProfileName

rubdos:


rubdos:
This seems okay and simple enough, merging as-is. CJKV-users will have their names in reverse order for now, so we should probably make that an issue.

PR #174, opened 2022-11-07
Use tokio spawn instead of spawn_local

gferon:
Otherwise, this fails at runtime depending on how you use the runtime.


rubdos:
```
     pub fn from_socket<WS: WebSocketService + 'static>(
         ws: WS,
         stream: WS::Stream,
-    ) -> (Self, std::pin::Pin<Box<dyn Future<Output = ()>>>) {
+    ) -> (Self, impl Future<Output = ()>) {
```

... ah that was the trick :-)

rubdos:
Ahh you know what, now again `libsignal-service-hyper` cannot compile with `--all-features`.

PR #176, opened 2022-11-08
Implement groupv2 avatar decryption

rubdos:
In progress of being tested out.


rubdos:
Easy enough and it works :-)

PR #178, opened 2022-11-10
More flags in Metadata

rubdos:
I want to store whether messages were received via sealed sending or sender keys, so I'm exposing these as flags in Metadata

### DIFF
--- a/libsignal-service/src/cipher.rs
+++ b/libsignal-service/src/cipher.rs
@@ -142,6 +142,7 @@ where
                     sender_device: envelope.source_device(),
                     timestamp: envelope.server_timestamp(),
                     needs_receipt: false,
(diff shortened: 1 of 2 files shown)


rubdos:
I'd like to expose a sender key flag too, but I don't think I can, without touching some branches in `sealed_sender_decrypt` :(

gferon:
Why do you need this?

rubdos:
> I'm curious: why do you need this?

To give some feedback in the UI. I'd like to add a small icon to show that something was sent sealed or not. Senderr keys is mostly for debugging purposes, I suppose.

rubdos:
image

Example of what I mean.

gferon (approved):
Sounds good!

PR #180, opened 2022-11-11
Implement profile avatar retention

rubdos:
Will test now.

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -14,6 +14,7 @@ use serde::{Deserialize, Serialize};
 use sha2::Sha256;
 use zkgroup::profiles::ProfileKey;
 
+use crate::push_service::AvatarWrite;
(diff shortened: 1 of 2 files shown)


rubdos:
This only [works after Rust 1.53](https://github.com/rust-lang/rust/pull/79278), so I'm desugaring the or-patterns for now.

rubdos:
This seems to work well.

gferon (approved):
LGTM! I'm only starting to think that we should think twice before adding more primitive arguments to our function and maybe introduce types so we can add variants later, but here it's perfectly fine 👍🏼🥳

rubdos:
> LGTM! I'm only starting to think that we should think twice before adding more primitive arguments to our function and maybe introduce types so we can add variants later, but here it's perfectly fine 👍🏼partying_face

I agree. In this case, it makes the methods very Rusty, I feel.

PR #182, opened 2022-11-13
Disable keep-alive on linking websocket

gferon:
This fixes linking as secondary device after refactoring of the web-socket code. The fix is a little bit lazy, as a refactoring of the `process` function could remove the interval clock entirely when `keep_alive` is `false`.


libsignal-service-hyper/src/provisioning.rs
```diff
@@ -45,7 +45,7 @@ pub async fn provision_secondary_device(
     );
 
     let mut push_service =
-        AwcPushService::new(service_configuration.clone(), None, USER_AGENT);
+        HyperPushService::new(service_configuration.clone(), None, USER_AGENT);
```
gferon, line 48: i also found this and was like __pikachu.jpg__

libsignal-service-actix/src/websocket.rs
```diff
@@ -69,6 +69,7 @@ impl From<AwcWebSocketError> for ServiceError {
 async fn process<S: Stream>(
     socket_stream: S,
     mut incoming_sink: Sender<WebSocketStreamItem>,
+    keep_alive: bool,
```
rubdos, line 72: I think I would've added this in the generic WebSocket, but I'm not against. This way, the events are not generated, and that's maybe even better.

rubdos (approved):

rubdos:
There's some CI failing, gferon, maybe you want to look at that first.

libsignal-service-actix/src/websocket.rs
gferon: Indeed, that's another approach, I kinda went quick and dirty with this one.

gferon: EDIT: the events _are_ generated in the implementation? which is why I did it here, see https://github.com/whisperfish/libsignal-service-rs/pull/182/files/32c11d40437badcfa5cc96d2788b54ddbaca9c33#diff-72f8f8faf866526c65b6bf1d14cc57b52f1c3b5e49f003245839bb4435a8de7eR92

gferon: the other way would be to match on requests, and discard the keep-alive?

rubdos: Totally fine.

rubdos (approved):
This is the API we want, modulo some compile-time or const generics. Let's get this in, if CI agrees :-)

PR #186, opened 2022-11-18
Reduce the memory size of MessageSenderError

boxdot:
The memory size of MessageSenderError is reduced from 128 to 64 bytes. Also add static assertions to maintain the size of the error.

### DIFF
--- a/libsignal-service/Cargo.toml
+++ b/libsignal-service/Cargo.toml
@@ -39,7 +39,6 @@ uuid = { version = "1", features = ["serde"] }
 phonenumber = "0.3"
 hkdf = "0.12"
 serde_json = "1.0.85"
-static_assertions = "1.1.0"
 
 [build-dependencies]
 prost-build = "0.10"
--- a/libsignal-service/Cargo.toml
+++ b/libsignal-service/Cargo.toml
@@ -39,6 +39,7 @@ uuid = { version = "1", features = ["serde"] }
 phonenumber = "0.3"
 hkdf = "0.12"
 serde_json = "1.0.85"
+static_assertions = "1.1.0"
 
 [build-dependencies]
 prost-build = "0.10"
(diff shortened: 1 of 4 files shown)


rubdos:
I'm definitely not against reducing the memory usage, but what's your reasoning that it necessarily needs to be 64 bytes?

boxdot:
> I'm definitely not against reducing the memory usage, but what's your reasoning that it necessarily needs to be 64 bytes?

The size is exactly at 64 bytes atm, and the error types from the libsignal-protocol are at 56 bytes. So, it will be hard to go below that because of the 8 bytes discriminant.

rubdos:
> The size is exactly at 64 bytes atm, and the error types from the libsignal-protocol are at 56 bytes. So, it will be hard to go below that because of the 8 bytes discriminant.

I understand the current size, but I don't understand why it needs to be asserted that it should not go over. If the error type gets expanded again, then the test fails, and I don't see what it guards against.

boxdot:
> > The size is exactly at 64 bytes atm, and the error types from the libsignal-protocol are at 56 bytes. So, it will be hard to go below that because of the 8 bytes discriminant.
> 
> I understand the current size, but I don't understand why it needs to be asserted that it should not go over. If the error type gets expanded again, then the test fails, and I don't see what it guards against.

Even 64 bytes for an error type is quite a lot. So, the idea was to guard going over it and in case this happens, make it visible for the developer that maybe they should Box their errors. However, I don't have a strong opinion about it, so I can also remove it.

boxdot:
> > Edit: btw. clippy warns when going over 128 bytes
> 
> Makes me wonder, can we put clippy at 80 bytes or something? That might make more sense.
> 
> I'm not against this, I was merely curious. I'll let @gferon decide :-)

Looks like it is indeed configurable (even the configuration does not work for me with rustc 1.65). Let me remove the static assertion.

PR #187, opened 2022-11-27
MessageSender to use the WS instead of a service

rubdos:
This seems to have been deprecated for a long time now, and it minimizes the amount of needed connections.


rubdos:
Sending works, fetching pre-keys does not yet.

rubdos:
Apparently pre-keys should remain via the push service.

rubdos:
This works and seems trivial enough to merge as-is. Let me know if the breakage is too much :-)

gferon:
> This works and seems trivial enough to merge as-is. Let me know if the breakage is too much :-)

If it works then 💯

rubdos:
> > This works and seems trivial enough to merge as-is. Let me know if the breakage is too much :-)
> 
> If it works then 100

It does, and it seems equivalent (apart from some error handling/fall back code) with the Android code too :-)

PR #188, opened 2022-11-28
Set avatar true if there's no request for deletion.

rubdos:
As far as the server code goes, this should work.

https://github.com/signalapp/Signal-Server/blob/main/service/src/main/java/org/whispersystems/textsecuregcm/entities/CreateProfileRequest.java#L97

@gferon would you mind double-checking my logic here?


gferon:
Looks correct to me, you could always expand this to a full blown `match` for readability?

PR #190, opened 2022-12-06
Handle HTTP status 428 when sending a message

### LINKED ISSUE #189: Send message HTTP status 428 not handled
Signal may respond to sending a message with a HTTP error 428. It indicates that the user must provide "proof" that they are not a robot. This has affected both new registrations sending the very first messages, and older registrations "out of the blue". A guesstimated list of tasks:

- [x] Catch the HTTP 428
- [x] Pass it to the client
- [x] Receive the completed proof from client
- [x] Send the proof to server
- [x] Tell the client if the proof was accepted

Related issues in Whisperfish:
- [Implement 428 response with a captcha](https://gitlab.com/whisperfish/whisperfish/-/issues/418)
- [Newly registered account gets 428 rate limiter errors](https://gitlab.com/whisperfish/whisperfish/-/issues/417)

direc85:
Fixes #189

### DIFF
--- a/libsignal-service-actix/src/push_service.rs
+++ b/libsignal-service-actix/src/push_service.rs
@@ -107,6 +107,15 @@ impl AwcPushService {
                 })?;
                 Err(ServiceError::StaleDevices(stale_devices))
             },
+            StatusCode::PRECONDITION_REQUIRED => {
+                let proof_required = response.json().await.map_err(|e| {
+                    log::error!("Failed to decode HTTP 428 response: {}", e);
+                    ServiceError::UnhandledResponseCode {
+                        http_code: StatusCode::PRECONDITION_REQUIRED.as_u16(),
+                    }
+                })?;
+                Err(ServiceError::ProofRequiredError(proof_required))
+            },
             // XXX: fill in rest from PushServiceSocket
             code => {
                 let contents = response.body().await;
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -109,6 +109,9 @@ pub enum MessageSenderError {
 
     #[error("Exceeded maximum number of retries")]
     MaximumRetriesLimitExceeded,
+
+    #[error("Proof of type {options:?} required using token {token}")]
+    ProofRequired { token: String, options: Vec<String> },
 }
 
 impl<Service, S, I, SP, P, SK, R> MessageSender<Service, S, I, SP, P, SK, R>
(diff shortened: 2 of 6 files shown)


libsignal-service/src/sender.rs
```diff
@@ -538,6 +541,15 @@ where
...
                     }
                 },
+                Err(ServiceError::ProofRequiredError(ref p)) => {
+                    log::debug!("{:?}", p);
+                    if p.options.contains(&String::from("recaptcha")) {
+                        return Err(MessageSenderError::ProofRequired {
+                            token: p.token.clone(),
+                            options: p.options.clone(),
+                        });
+                    }
```
rubdos, line 551: What about the `else` case, then it will just return an `Ok` and silently fail?

direc85:
I believe this is "barely complete enough" to receive proper review. The basic functionality _should_ be there, but because I haven't been hitting the 428 error myself, and submitting the rate limit reCaptcha response requires the token received with the 428 error, this is effectively untested.

libsignal-service-actix/src/push_service.rs
```diff
@@ -107,6 +107,15 @@ impl AwcPushService {
...
             },
+            StatusCode::PRECONDITION_REQUIRED => {
+                let proof_required = response.json().await.map_err(|e| {
+                    log::error!("Failed to decode HTTP 428 response: {}", e);
+                    ServiceError::UnhandledResponseCode {
+                        http_code: StatusCode::PRECONDITION_REQUIRED.as_u16(),
+                    }
+                })?;
+                Err(ServiceError::ProofRequiredError(proof_required))
+            },
```
rubdos, line 118: We might also want this in the websocket, which requires #191. I don't think this code actually works on Whisperfish, after we started using the WS for sending messages?

direc85: I think this is now fixed in [81caabe](https://github.com/whisperfish/libsignal-service-rs/pull/190/commits/81caabe071e11ff3c51c66bb27dbe1fee9a6b5e3)..?

(I still plan to squish the commits for cleaner commit history.)

libsignal-service/src/sender.rs
direc85: Fixed by not filtering the captcha type here; let the client do it instead.

rubdos: Sounds good to me, we can always refine that later.

libsignal-service-actix/src/push_service.rs
direc85: It's [ca14821](https://github.com/whisperfish/libsignal-service-rs/pull/190/commits/ca14821f02a7371a7bd472bcb436fe4d1a7c0ac1) after cleanup.

PR #191, opened 2022-12-18
Handle errors (409/410) in WS json request

rubdos:
I'm not sure where this happens in Signal-Android, but it seems like this might be a good place to put this code. @gferon, would you maybe have time to take a look? @direc85 you could test whether this actually solves your unhandled 409's.

### DIFF
--- a/libsignal-service/src/websocket.rs
+++ b/libsignal-service/src/websocket.rs
@@ -17,7 +17,7 @@ use crate::proto::{
     web_socket_message, WebSocketMessage, WebSocketRequestMessage,
     WebSocketResponseMessage,
 };
-use crate::push_service::ServiceError;
+use crate::push_service::{MismatchedDevices, ServiceError};
 
 mod attachment_service;
 mod sender;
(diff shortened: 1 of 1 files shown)


rubdos:
The reason I'm not sure that this is the best place, is because different endpoints could potentially have different meanings for different errors. That would mean that the 409/410 handling should go in its respective call.

gferon:
I can check tomorrow but it might too late depending on your timeline 🏎️

direc85:
Instead of a crash, I'm now getting an error message, so I guess it works.

I think it's fine to merge this as-is, and split the handling across relevant calls when the need arises.

libsignal-service/src/websocket.rs
```diff
@@ -377,19 +377,61 @@ impl SignalWebSocket {
...
+        {
+            serde_json::from_slice(body).map_err(|e| {
+                ServiceError::JsonDecodeError {
+                    reason: e.to_string(),
+                }
+            })
+        }
+
+        match response.status() {
+            200 => json(response.body()),
```
gferon, line 398: nit: `200 | 204`

libsignal-service/src/websocket.rs
```diff
@@ -377,19 +377,61 @@ impl SignalWebSocket {
...
+            204 => json(response.body()),
+            401 => Err(ServiceError::Unauthorized),
```
gferon, line 400: same here

gferon (approved):
LGTM!

libsignal-service/src/websocket.rs
rubdos: I didn't dare, because I feared the `200 | 204 == 204` because of bitwise or. Do you know which of both options it takes?

rubdos: The left part of each arm of a match expression is a pattern not an expression. That means you can only do pattern matching there. You can't use operators, accessors and function calls.

Will change.

PR #192, opened 2023-01-09
Make group types ser/de

gferon:


rubdos:
Out of curiosity, what is that for? I would think you'd rather store these as protobuf encoded, if it's for local storage.

gferon:
This is exactly the idea, yes. Most of the `group` stuff is mirrored with our own internal types, mostly because the protobuf definition contains a bunch of byte arrays that we decrypt and validate ourselves (and then convert to a usable representation). Once you've done this work, it would be a shame to go back to a lower level type.

rubdos:
Then, Rust-wise speaking, wouldn't it be a better idea to convert to our own internal type? If  Signal changes the struct definition, the protobuf will keep working, but (y)our JSON serialization can trivially break.

With our own types (like my ORM module in Whisperfish), this generates a compile error instead of potentially failing at runtime or going unnoticed.

PR #193, opened 2023-01-29
Split GroupsManager::get_group into two separate functions

gferon:
I need this to be able to store the encrypted group data in the local DB of `presage`. This is also a slight improvement because we can hide a bunch of implementation details (no need to know how to build a `group_secret_params` and `credentials`).

### DIFF
--- a/libsignal-service/src/groups_v2/manager.rs
+++ b/libsignal-service/src/groups_v2/manager.rs
@@ -108,19 +108,22 @@ impl CredentialsCache for InMemoryCredentialsCache {
     }
 }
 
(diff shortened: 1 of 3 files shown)


rubdos (approved):
Seems to make sense to me. You need to store the encrypted state in order to fix out-of-order things?

gferon:
> Seems to make sense to me. You need to store the encrypted state in order to fix out-of-order things?

I mostly need this because the easiest way to add groups persistence in presage is to store the master-key encrypted protobuf representation of groups metadata.

PR #194, opened 2023-01-30
Update protobufs

gferon:
and add a script to automate the job.

### DIFF
--- a/libsignal-service/build.rs
+++ b/libsignal-service/build.rs
@@ -28,5 +28,9 @@ fn main() {
         })
         .collect();
 
-    prost_build::compile_protos(&input, &[protobuf]).unwrap();
+    let mut prost_build = prost_build::Config::new();
+    prost_build
+        .default_package_filename("signal")
+        .compile_protos(&input, &[protobuf])
+        .unwrap();
 }
--- a/libsignal-service/src/cipher.rs
+++ b/libsignal-service/src/cipher.rs
@@ -79,12 +79,7 @@ where
         &mut self,
         envelope: Envelope,
     ) -> Result<Option<Content>, ServiceError> {
-        if envelope.legacy_message.is_some() {
-            let plaintext = self.decrypt(&envelope).await?;
-            let message =
-                crate::proto::DataMessage::decode(plaintext.data.as_slice())?;
-            Ok(Some(Content::from_body(message, plaintext.metadata)))
-        } else if envelope.content.is_some() {
+        if envelope.content.is_some() {
             let plaintext = self.decrypt(&envelope).await?;
             let message =
                 crate::proto::Content::decode(plaintext.data.as_slice())?;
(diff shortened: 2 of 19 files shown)


libsignal-service/build.rs
```diff
@@ -28,5 +28,9 @@ fn main() {
         })
         .collect();
 
-    prost_build::compile_protos(&input, &[protobuf]).unwrap();
+    let mut prost_build = prost_build::Config::new();
+    prost_build
+        .default_package_filename("signal")
```
rubdos, line 33: Does this break compat?

libsignal-service/src/cipher.rs
```diff
@@ -79,12 +79,7 @@ where
         &mut self,
         envelope: Envelope,
     ) -> Result<Option<Content>, ServiceError> {
-        if envelope.legacy_message.is_some() {
-            let plaintext = self.decrypt(&envelope).await?;
-            let message =
-                crate::proto::DataMessage::decode(plaintext.data.as_slice())?;
-            Ok(Some(Content::from_body(message, plaintext.metadata)))
-        } else if envelope.content.is_some() {
+        if envelope.content.is_some() {
```
rubdos, line 82: We can probably drop the whole envelope decryption system. It's been unused for ages now anyway.

libsignal-service/src/envelope.rs
```diff
@@ -129,9 +128,7 @@ impl Envelope {
             source_device: Some(entity.source_device),
             timestamp: Some(entity.timestamp),
             server_timestamp: Some(entity.server_timestamp),
-            source_e164: source.e164(),
```
rubdos, line 132: This scares me. A lot.

libsignal-service/src/envelope.rs
```diff
@@ -154,23 +151,16 @@ impl Envelope {
...
         let uuid = self
             .source_uuid
             .as_deref()
             .map(Uuid::parse_str)
             .transpose()
             .expect("valid e164 checked in constructor");
+
         ServiceAddress {
-            phonenumber,
+            phonenumber: None,
```
rubdos, line 162: Even more scary. Should we modify `ServiceAddress` to be a wrapper around `Uuid`?

rubdos (approved):
Some minor questions, but if the questions make sense, then the PR makes sense too.

libsignal-service/src/envelope.rs
gferon: Well, it's really not part of the proto anymore, so it can technically be sent, but can't be used in the official clients anymore.

libsignal-service/src/envelope.rs
gferon: It's still in use in other parts of the code IIRC, so not all code paths are e164-less yet. I'll double check. (I think it was unidentified message somewhere).

libsignal-service/src/envelope.rs
rubdos: Oh it's totally fine. Just scary.

libsignal-service/src/envelope.rs
rubdos: Let's make it an issue that we remove the phone number field, if that's the same in Android/iOS.

libsignal-service/build.rs
gferon: Quite the opposite, this is there to keep compatibility 🪨

libsignal-service/src/cipher.rs
gferon: Which part do you mean exactly? `ServiceCipher::decrypt` is where all the magic happens IIRC and we call it here.

PR #195, opened 2023-02-06
Remove our own ProfileKey type and expose the zkgroup one

gferon:
Trying to remove inconsistencies with types from `libsignal` and our own. We only used this one for the `.derive_access_key` method, which I moved into a `ProfileKeyExt` trait instead.

### DIFF
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -1,4 +1,4 @@
-use std::{convert::TryInto, fmt, ops::Deref, time::Duration};
+use std::{fmt, time::Duration};
 
 use crate::{
     configuration::{Endpoint, ServiceCredentials},
@@ -24,7 +24,7 @@ use libsignal_protocol::{
 use prost::Message as ProtobufMessage;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
-use zkgroup::profiles::{ProfileKeyCommitment, ProfileKeyVersion};
+use zkgroup::profiles::{ProfileKey, ProfileKeyCommitment, ProfileKeyVersion};
 
 /**
 Since we can't use format!() with constants, the URLs here are just for reference purposes
@@ -135,9 +135,6 @@ pub struct ProofRequired {
     pub options: Vec<String>,
 }
 
-#[derive(Clone)]
-pub struct ProfileKey(pub [u8; 32]);
-
 #[derive(Debug, Deserialize, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct PreKeyStatus {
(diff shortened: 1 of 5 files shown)


libsignal-service/src/push_service.rs
```diff
@@ -170,9 +167,13 @@ impl fmt::Debug for HttpAuth {
...
-impl ProfileKey {
-    pub fn derive_access_key(&self) -> Vec<u8> {
-        let key = GenericArray::from_slice(&self.0);
+pub trait ProfileKeyExt {
+    fn derive_access_key(&self) -> Vec<u8>;
+}
+
+impl ProfileKeyExt for ProfileKey {
+    fn derive_access_key(&self) -> Vec<u8> {
+        let key = GenericArray::from_slice(&self.bytes);
```
rubdos, line 176: Should we make a PR to zkgroup for this?

libsignal-service/src/websocket.rs
```diff
@@ -104,11 +104,12 @@ impl<WS: WebSocketService> SignalWebSocketProcess<WS> {
...
-                self.request_sink.send((request, sink)).await.map_err(
-                    |_| ServiceError::WsError {
+                self.request_sink.send((request, sink)).await.map_err(|e| {
+                    log::error!("websocket request failed: {e}");
+                    ServiceError::WsError {
                         reason: "request handler failed".into(),
-                    },
-                )?;
+                    }
+                })?;
```
rubdos, line 112: This seems irrelevant for this PR.

libsignal-service/src/websocket.rs
```diff
@@ -291,10 +292,11 @@ impl SignalWebSocket {
             ws,
             stream,
         };
+
         let process = process.run().map(|x| match x {
             Ok(()) => (),
-            Err(e) => {
-                log::error!("SignalWebSocket: {}", e);
+            Err(error) => {
+                log::error!("{error}");
```
rubdos, line 299: Same, irrelevant to this PR.

rubdos (changes requested):

libsignal-service/src/push_service.rs
gferon: I think so, but I always try to avoid it 😅

gferon:
@rubdos sorry about the unrelated changes that were me trying to find a heisenbug, this is good to go now!

rubdos:
> EDIT: whisperfish doesn't seem to use these types to store the data, so I'm 99.99% sure things won't break.

Don't worry about breaking Whisperfish. I care about getting this library is good shape :)

libsignal-service/src/push_service.rs
rubdos: I'll give it a try.

rubdos (approved):

rubdos:
By the way, no need to wait for signalapp/libsignal#508, that's going to take some time anyway!

Issue #65, opened 2021-01-31
Refactor get_preferred_protocol_address


rubdos:
Things like bd7f9eeab86ec74b1d46c053511afd6839f60ae1 slip through reviews often.

1. I think we might want to have `get_preferred_protocol_address()` as a method on either `Address` (as constructor), on `StorageContext` (as provided method) or as method on `ServiceAddress` (as conversion method into an Address).
2. I think we may want `StoreContext::get_sub_device_sessions` to accept a ServiceAddress instead of an `Address`, or make a provided method on the trait.

Both these items correspond to an `XXX` in the above commit.

PR #198, opened 2023-02-08
Remove E164 phone numbers from ServiceAddress

### LINKED ISSUE #196: Remove e164 / phone number from ServiceAddress
And possibly alias the type to `Uuid` entirely.

gferon:
TODO:

- [x] change types everywhere we can to not use `&str` since we now have a single identifier

Closes #196 #65

### DIFF
--- a/libsignal-service/src/cipher.rs
+++ b/libsignal-service/src/cipher.rs
@@ -414,12 +414,12 @@ pub async fn get_preferred_protocol_address<S: SessionStore>(
     address: &ServiceAddress,
     device_id: DeviceId,
 ) -> Result<ProtocolAddress, libsignal_protocol::error::SignalProtocolError> {
-    let address = ProtocolAddress::new(address.to_string(), device_id);
+    let address = address.to_protocol_address(device_id);
     if session_store.load_session(&address, None).await?.is_some() {
         return Ok(address);
     }
 
-    Ok(ProtocolAddress::new(address.to_string(), device_id))
+    Ok(address)
 }
 
 /// Decrypt a Sealed Sender message `ciphertext` in either the v1 or v2 format, validate its sender
--- a/libsignal-service/src/cipher.rs
+++ b/libsignal-service/src/cipher.rs
@@ -206,7 +206,7 @@ where
             Type::UnidentifiedSender => {
                 let SealedSenderDecryptionResult {
                     sender_uuid,
-                    sender_e164,
+                    sender_e164: _,
                     device_id,
                     mut message,
                 } = sealed_sender_decrypt(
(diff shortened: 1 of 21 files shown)


libsignal-service/src/cipher.rs
```diff
@@ -206,7 +206,7 @@ where
             Type::UnidentifiedSender => {
                 let SealedSenderDecryptionResult {
                     sender_uuid,
-                    sender_e164,
+                    sender_e164: _,
```
rubdos, line 209: Something tells me that this is also removed with a protobuf update?

libsignal-service/src/service_address.rs
```diff
@@ -1,124 +1,64 @@
...
-    InvalidPhoneNumber(#[from] phonenumber::ParseError),
-
-    #[error("Supplied uuid could not be parsed")]
-    InvalidUuidError(#[from] uuid::Error),
+    #[error("Supplied UUID could not be parsed")]
+    InvalidUuid(#[from] uuid::Error),
 
-    #[error("Envelope with neither Uuid or E164")]
-    NoSenderError,
+    #[error("Envelope with no Uuid")]
```
rubdos, line 11:
```suggestion
    #[error("Envelope without Uuid")]
```

libsignal-service/src/push_service.rs
```diff
@@ -720,9 +718,9 @@ pub trait PushService: MaybeSend {
         device_id: u32,
     ) -> Result<Vec<PreKeyBundle>, ServiceError> {
         let path = if device_id == 1 {
-            format!("/v2/keys/{}/*", destination.identifier())
+            format!("/v2/keys/{}/*", destination)
         } else {
-            format!("/v2/keys/{}/{}", destination.identifier(), device_id)
+            format!("/v2/keys/{}/{}", destination, device_id)
```
rubdos, line 723: I would explicitely ask the destination.uuid here. Display is for user-friendly representation, not necessarily machine-useful representation. See below.

libsignal-service/src/service_address.rs
```diff
@@ -1,124 +1,64 @@
...
-            );
-        }
-        if let Some(e164) = &self.e164() {
-            sub_device_sessions
-                .extend(session_store.get_sub_device_sessions(e164).await?);
-        }
-        Ok(sub_device_sessions)
+impl fmt::Display for ServiceAddress {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.uuid)
```
rubdos, line 22: I'm not sure whether I necessarily agree with that. I think I'd rather see ServiceAddress(uuid) then.

This implies you would rely on this specific Display implementation for certain calls (see above) and that doesn't sound very healthy.

libsignal-service/src/websocket/sender.rs
```diff
@@ -3,11 +3,11 @@ use crate::sender::{OutgoingPushMessages, SendMessageResponse};
...
 
 impl SignalWebSocket {
-    pub async fn send_messages<'a>(
+    pub async fn send_messages(
         &mut self,
-        messages: OutgoingPushMessages<'a>,
+        messages: OutgoingPushMessages,
     ) -> Result<SendMessageResponse, ServiceError> {
-        let path = format!("/v1/messages/{}", messages.destination);
+        let path = format!("/v1/messages/{}", messages.recipient);
```
rubdos, line 10: Same here, try not to depend on the Display implementation of a struct that should be an opaque type. Explicitely request the uuid here.

rubdos (changes requested):
Some minor remarks.

libsignal-service/src/cipher.rs
gferon: This is a [concrete type from `libsignal`](https://github.com/signalapp/libsignal/blob/main/rust/protocol/src/sealed_sender.rs#L158) which has the phone number as `Option<String>`.

libsignal-service/src/push_service.rs
gferon: I have remove the `impl fmt::Display` altogether, as I think it can be misused and we don't really need to ever print the `ServiceAddress` as-is.

rubdos (approved):
We still have a Debug impl, so that should cut all the other cases where you want to "display" the ServiceAddress.

PR #201, opened 2023-02-22
Use string instead of u32 for confirmation_code

nanu-c:
The `confirmation_code` should be a string because else when the code contains a `0` as first digit this will be stripped.


gferon:
What about taking `impl AsRef<str>` instead?

gferon (approved):

gferon (commented):
Use impl AsRef<str>

PR #208, opened 2023-03-19
Add group sync facilities

gferon:


libsignal-service/src/receiver.rs
```diff
@@ -52,58 +58,77 @@ impl<Service: PushService> MessageReceiver<Service> {
...
+        let key_material = attachment.key();
+        assert_eq!(
+            key_material.len(),
+            64,
+            "key material for attachments is ought to be 64 bytes"
+        );
+        let mut key = [0u8; 64];
+        key.copy_from_slice(key_material);
+
+        decrypt_in_place(key, &mut ciphertext).expect("attachment decryption");
```
rubdos, line 101: This sounds like an error we want to return instead of crash on.

gferon:
Turns out this is not needed, as it is only for groups v1 (the primary device still sends an attachment after linking, but it contains padded zero bytes 😄)

PR #211, opened 2023-03-30
Blanket impl for mutable references to credentialcache

rubdos:
Before, the GroupManager took `&mut CredentialCache`, but I cannot move my impl into the GroupManager, so I need a blanket `impl CredentialCache for &mut CredentialCache` :-)


rubdos:
(does this make sense to implement? could this break someone else's generic impls?)

gferon (approved):
LGTM!

PR #215, opened 2023-03-30
Propagate NotFoundError

direc85:
When getting prekeys, getting a 404 reply means that the recipient is unregistered. This needs to be handled in the application side, so let's propagate this error out of libsignal-service.

Required by https://gitlab.com/whisperfish/whisperfish/-/merge_requests/396 which doesn't compile against master, but it soon should!

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -734,17 +734,12 @@ where
                     log::trace!("Get prekeys OK");
                     ok
                 },
-                Err(e) => {
-                    log::trace!("Get prekeys failed: {}", e);
-                    return match e {
-                        ServiceError::NotFoundError => {
-                            Err(MessageSenderError::NotFound {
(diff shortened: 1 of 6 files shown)


libsignal-service/src/sender.rs
```diff
@@ -709,10 +725,27 @@ where
...
+                    return match e {
+                        ServiceError::NotFoundError => {
+                            Err(MessageSenderError::NotFound {
+                                uuid: recipient.uuid,
+                            })
+                        },
+                        _ => Err(From::from(e)),
+                    };
+                },
+            };
```
rubdos, line 748:
```suggestion
                Err(ServiceError::NotFoundError) => {
                    return Err(MessageSenderError::NotFound {
                        uuid: recipient.uuid,
                    });
                },
                Err(e) => Err(e)?,
            };
```

PR #217, opened 2023-04-04
Make groups struct ser/de

gferon:
This is a remake of #192 which makes we use 100% internal types when dealing with group decryption. Before this change, we had a mix of own types and protobuf enums. The main advantage of using our own types is to be able to control derives as well as immediately catch issues when the protobuf schema changes.

- Move all group structs and enums into its own `model.rs` file
- Add 3 very small converters for `Role`, `AccessRequired` and `AccessControl`.

### DIFF
--- a/libsignal-service/src/groups_v2/model.rs
+++ b/libsignal-service/src/groups_v2/model.rs
@@ -5,7 +5,7 @@ use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 use zkgroup::profiles::ProfileKey;
 
-use super::GroupDecryptionError;
+use super::GroupDecodingError;
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub enum Role {
@@ -139,7 +139,7 @@ pub struct Timer {
 /// Conversion from protobuf definitions
 
 impl TryFrom<i32> for Role {
-    type Error = GroupDecryptionError;
+    type Error = GroupDecodingError;
 
     fn try_from(value: i32) -> Result<Self, Self::Error> {
         use crate::proto::member::Role::*;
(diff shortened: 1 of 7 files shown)


libsignal-service/src/groups_v2/model.rs
```diff
@@ -0,0 +1,246 @@
...
+    AttributeAccess(AccessRequired),
+    MemberAccess(AccessRequired),
+    InviteLinkAccess(AccessRequired),
+    InviteLinkPassword(String),
+    AnnouncementOnly(bool),
+}
+
+impl fmt::Debug for GroupChange {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
```
rubdos, line 135: What's the reason for not deriving here? If there's a legitimate reason, maybe document it as a comment, because I don't see it at first sight.

libsignal-service/src/groups_v2/model.rs
```diff
@@ -0,0 +1,246 @@
...
+            Some(Member) => Ok(AccessRequired::Member),
+            Some(Administrator) => Ok(AccessRequired::Administrator),
+            Some(Unsatisfiable) => Ok(AccessRequired::Unsatisfiable),
+            None => Err(GroupDecryptionError::WrongEnumValue),
+        }
+    }
+}
+
+impl TryFrom<crate::proto::AccessControl> for AccessControl {
+    type Error = GroupDecryptionError;
```
rubdos, line 235: I feel like this might have to be a `ServiceError` instead?

rubdos (changes requested):
Other than those two comments, looks very neat to me. I was first wondering whether these should've been structs in the client libraries instead, but it does make sense, I think. At Whisperfish' side, we'll implement our own serialization either way (through Diesel).

libsignal-service/src/groups_v2/model.rs
gferon: 🍝

libsignal-service/src/groups_v2/model.rs
gferon: Done!

PR #218, opened 2023-04-05
Publish group operation structs

Schmiddiii:
Needed for <https://github.com/whisperfish/presage/pull/149>. Note that the base-branch is quite outdated as presage was not yet updated to latest commit.


rubdos:
Makes sense.

PR #220, opened 2023-04-13
Bump actix to 0.13

rubdos:
This should work still on 1.52


rubdos:
My build over at Whisperfish seems fine with this, don't see any extra duplicate deps. Merging :)

Issue #206, opened 2023-03-09
Add support for ACI/PNI


gferon:
Since we don't support PNI yet, we might want to ACK the messages we receive to avoid seeing them in the loop all the time.

Maybe this is a bad idea and we should spend effort implementing it instead.

gferon:
Note: basic support for making sure we have a PNI UUID is in place, but we currently hardcode the use of `ServiceIdType::AccountIdentity` in a bunch of helper methods like `update_pre_key_bundle`. Fully supporting PNI requires having two separate implementations of the `ProtocolStore` that end up storing a separate key-pair, sessions as well as pre-keys.

@rubdos I think a good way to do this would be to add our own `ProtocolStore` trait (or `AccountStore`?) that has two methods: `aci_protocol_store` and `pni_protocol_store`. That way, we _could_ have functions only take a `ServiceIdType` as a parameter, and mimick what is done in Signal-Android. Design wise, we can really think of whatever works best (and helps us avoid adding more type parameters).

PR #224, opened 2023-04-22
Add support for PNI fields and fix fetching (encrypted groups)

### LINKED ISSUE #223: Groups v2 getAuthorizationForToday stopped working
I'll start investigating now, but I believe it might be related to PNI 😓

gferon:
Tested in `presage` which should get us started with #206 and fix #223.

This requires to be able to provide two separate store implementations to the underlying `libsignal`. See https://github.com/signalapp/Signal-Android/blob/b4f6177e873cfd23cbce8fdbea8b573af8faf45d/app/src/main/java/org/thoughtcrime/securesms/crypto/storage/SignalServiceAccountDataStoreImpl.java#L26 for the requirements in `Signal-Android`.

### DIFF
--- a/libsignal-service/src/provisioning/manager.rs
+++ b/libsignal-service/src/provisioning/manager.rs
@@ -15,7 +15,7 @@ use crate::{
     configuration::{Endpoint, ServiceCredentials, SignalingKey},
     push_service::{
         AccountAttributes, DeviceId, HttpAuthOverride, PushService,
-        ServiceError,
+        ServiceError, ServiceIds,
     },
     utils::serde_base64,
 };
@@ -50,7 +50,7 @@ pub enum VerificationCodeResponse {
 #[serde(rename_all = "camelCase")]
 pub struct VerifyAccountResponse {
     pub uuid: Uuid,
-    pub pni: String,
+    pub pni: Uuid,
     pub storage_capable: bool,
 }
 
@@ -68,10 +68,13 @@ pub enum SecondaryDeviceProvisioning {
         phone_number: phonenumber::PhoneNumber,
         device_id: DeviceId,
         registration_id: u32,
-        uuid: Uuid,
+        service_ids: ServiceIds,
         #[derivative(Debug = "ignore")]
-        private_key: PrivateKey,
-        public_key: PublicKey,
+        aci_private_key: PrivateKey,
+        aci_public_key: PublicKey,
+        #[derivative(Debug = "ignore")]
+        pni_private_key: PrivateKey,
+        pni_public_key: PublicKey,
         #[derivative(Debug = "ignore")]
         profile_key: Vec<u8>,
     },
@@ -258,7 +261,7 @@ impl<P: PushService> LinkingManager<P> {
                         .expect("failed to send provisioning Url in channel");
                 },
                 Ok(ProvisioningStep::Message(message)) => {
-                    let uuid = message
+                    let aci_uuid = message
                         .aci
                         .ok_or(ProvisioningError::InvalidData {
                             reason: "missing client UUID".into(),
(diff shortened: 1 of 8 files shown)


gferon:
@rubdos did you investigate by any chance how one can "upgrade" an existing account and upload some PNI keypair? I don't care that much for linking, as one can simply ask to relink (without losing any data).

libsignal-service/src/provisioning/manager.rs
```diff
@@ -271,22 +274,51 @@ impl<P: PushService> LinkingManager<P> {
...
                         )?,
                     )?;
 
+                    let pni_public_key = PublicKey::deserialize(
+                        &message.pni_identity_key_public.ok_or(
+                            ProvisioningError::InvalidData {
+                                reason: "missing public key".into(),
+                            },
+                        )?,
+                    )?;
```
rubdos, line 312: Should this indeed be a hard error? I imagine PNI is optional?

rubdos (commented):
I have not looked into the migration logic here, no. I would rather actually implement the migration. It's not because relinking is supported in-protocol, that we can do that easily in Whisperfish already :see_no_evil:

libsignal-service/src/provisioning/manager.rs
gferon: Yeah, this is what I thought. I can try to look into it.

gferon: No, it's always there.

rubdos: Makes sense during linking. Mergable.

gferon:
> I have not looked into the migration logic here, no. I would rather actually implement the migration. It's not because relinking is supported in-protocol, that we can do that easily in Whisperfish already :see_no_evil: 

Yes this is what I figured. I can look into the migration, i suppose you can upload a PNI private and public key, and get some UUID back.

libsignal-service/src/push_service.rs
```diff
@@ -67,8 +67,26 @@ pub const STICKER_PATH: &str = "stickers/%s/full/%d";
 pub const KEEPALIVE_TIMEOUT_SECONDS: Duration = Duration::from_secs(55);
 pub const DEFAULT_DEVICE_ID: u32 = 1;
 
+#[derive(Debug)]
+pub enum ServiceIdType {
+    // Account Identity
+    ACI,
+    // Phone number identity
+    PNI,
+}
```
rubdos, line 76:
```suggestion
#[derive(Debug)]
pub enum ServiceIdType {
    // Account Identity
    Aci,
    // Phone number identity
    Pni,
}
```

That's more rusty...

libsignal-service/src/push_service.rs
```diff
@@ -69,17 +69,17 @@ pub const DEFAULT_DEVICE_ID: u32 = 1;
...
 #[derive(Debug)]
 pub enum ServiceIdType {
-    // Account Identity
-    ACI,
-    // Phone number identity
-    PNI,
+    /// Account Identity (an account UUID without an associated phone number, probably in the future to a username)
+    AccountIdentity,
+    /// Phone number identity (a UUID associated with a phone number)
+    PhoneNumberIdentity,
```
rubdos, line 75:
```suggestion
    /// Account Identity (ACI)
    ///
    /// An account UUID without an associated phone number, probably in the future to a username
    AccountIdentity,
    /// Phone number identity (PNI)
    ///
    /// A UUID associated with a phone number
    PhoneNumberIdentity,
```

rubdos (commented):
Nitpick on the docstrings for the ServiceIdType.

To what extend did you look at the migration, yet?

gferon:
> To what extend did you look at the migration, yet?

To repeat what was said on Matrix, it looks like the migration is as easy as registering pre-keys with the server for `ServiceIdType::PhoneNumberIdentity`. We should be able to merge this without changing existing behaviour, and it'll let me experiment with `presage`.

PR #225, opened 2023-04-23
Avoid JsonDecodeError with empty 200 OK response

direc85:
Not sure if this is the correct way to handle this, but since [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.1) allows zero-length response for 200 OK, we should be able to handle.

Basically, don't try to `serde_json` an empty source, only for it to certainly fail. Generate an empty JSON object instead, which should be close enough to "empty payload"...


libsignal-service-actix/src/push_service.rs
```diff
@@ -170,31 +170,35 @@ impl PushService for AwcPushService {
...
-        } else {
-            response
-                .json()
-                .await
-                .map_err(|e| ServiceError::JsonDecodeError {
-                    reason: e.to_string(),
+        match response.body().await {
+            Ok(text) if text.len() == 0 => {
+                log::debug!("Empty GET response");
+                serde_json::from_slice(b"null").map_err(|e| {
```
gferon, line 182: Maybe we could simplify the `match` by returning the `&str` (so either `text` or `"null"`) and have the `serde_json::from_slice` block a single time? I would also look into refactoring errors that store strings, but that's for another PR I think.

gferon (changes requested):
Very nice, thanks! One minor comment, otherwise LGTM.

libsignal-service-actix/src/push_service.rs
direc85: It was simpler to return the bytes or error out, and check the length in `from_slice()` instead, but the result is a lot cleaner now! I also opted to return `ServiceError::ResponseError` since it's returned in similar places in hyper.

rubdos:
[ee8cd08](https://github.com/whisperfish/libsignal-service-rs/pull/225/commits/ee8cd084b6509e85601d404b71fee487a259ae92) cleans up some redundancy and removes the need for the parsing of a `&'static str`.

libsignal-service-actix/src/push_service.rs
rubdos: I have refactored this a bit more, let's ship this.

PR #229, opened 2023-05-13
Add pni_registration_id to linking manager

rubdos:
While technically I guess this could be registered after the fact with an account attribute update, I think this is the better approach (esp. when Signal will make it mandatory during linking).

### DIFF
--- a/libsignal-service/src/provisioning/manager.rs
+++ b/libsignal-service/src/provisioning/manager.rs
@@ -29,6 +29,7 @@ pub(crate) struct ConfirmDeviceMessage {
     pub supports_sms: bool,
     pub fetches_messages: bool,
     pub registration_id: u32,
+    pub pni_registration_id: u32,
     #[serde(with = "serde_base64", skip_serializing_if = "Vec::is_empty")]
     pub name: Vec<u8>,
 }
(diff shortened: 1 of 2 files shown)


rubdos:
Very much untested, just bumped on it during assumptions and type checking :'-)

libsignal-service/src/provisioning/manager.rs
```diff
@@ -248,6 +250,7 @@ impl<P: PushService> LinkingManager<P> {
             .await?;
 
         let registration_id = csprng.gen_range(1, 256);
+        let pni_registration_id = csprng.gen_range(1, 256);
```
gferon, line 253: I actually wanted to test it but there's a method we use from `libsignal` that does this in a wider range, and I think it's like this because it's inconsistent (or old). But maybe we can change this in another PR.

gferon:
LGTM (compilation issues aside)

libsignal-service/src/provisioning/manager.rs
rubdos: Yeh, I think it needs to be until somewhere in the 16k...

rubdos:
> compilation issues aside

Thanks, didn't notice.

PR #231, opened 2023-06-01
CI: Add grcov coverage analysis

rubdos:


rubdos:
Looks like this will have worked too https://app.codecov.io/github/whisperfish/libsignal-service-rs/commit/c54356e1033e3b32b01853448cc7844bb6bad78b

PR #232, opened 2023-06-07
Add api for retrieving stickerpacks from the cdn

R-D-C-N:
While mostly untested, this should allow for retrieval of stickerpack data by requesting from the correct endpoints using the preexisting cdn retrieval functions.

### DIFF
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -592,6 +592,23 @@ pub trait PushService: MaybeSend {
         }
     }
 
+    async fn get_sticker_pack_manifest(
+        &mut self,
+        id: &str,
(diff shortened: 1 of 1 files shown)


rubdos (approved):
LGTM, although maybe we want to revise/amend the API to actually decrypt the ByteStreams that are yielded here.

libsignal-service/src/push_service.rs
```diff
@@ -592,6 +592,23 @@ pub trait PushService: MaybeSend {
         }
     }
 
+    async fn get_sticker_pack_manifest(
+        &mut self,
+        id: &str
```
rubdos, line 597:
```suggestion
        id: &str,
```

rustfmt has two comments

rubdos (approved):

PR #233, opened 2023-06-16
Add a function for deriving the sticker pack key

R-D-C-N:


rubdos:
Sounds like this should become a whole sticker decryption module, instead of just a key derivation function.

I suggest that we architect such a function in this PR. If you don't feel up to it, I would leave the PR open such that someone (potentially myself) can pick it up.

R-D-C-N:
From what I have gathered from how sticker and manifest decryption/encryption is implemented: It is identical to attachment encryption with one difference, that difference being that the cryptographic key needs to be derived from the pack key.

Continuing along the lines of your comment, if you think there should be additional separate functions that just call the attachment function or some kind of structure that handles the decryption for a pack, that could be added to my PR.

Sorry for not mentioning this earlier.

PR #234, opened 2023-07-17
Post-quantum E2E sessions

rubdos:
Since two months, Signal supports PQ E2E sessions. This PR adds this support to libsignal-service-rs.

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -207,6 +207,12 @@ impl<Service: PushService> AccountManager<Service> {
             pq_pre_keys: pq_pre_key_entities,
             pq_last_resort_key: if use_last_resort_key {
                 log::warn!("Last resort Kyber key unimplemented");
+                // Note about the last-resort key:
+                // mark_kyber_pre_key_used() should retain the last-resort key, but can safely
+                // remove the ephemeral pre keys.  This implies that generating the last-resort key
+                // should notify the pre-key store, when saving the key, that it concerns a
+                // last-resort key.  I don't see how this can be communicated to the store, and I
+                // fear that we need to reengineer the whole prekeystore system as a whole.
                 None
                 // Some(KyberPreKeyEntity {
                 //     key_id: 0x7fffffff,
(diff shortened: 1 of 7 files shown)


shekohex (commented):
Looks good, there is other few files like the hyper specific crate  has now a few errors, maybe due to this upgrade, so lets fix that  too.

rubdos:
Registering Kyber keys with the server seems to work correctly. However, I don't have any contact that seems to want to establish a Kyber session yet.

rubdos:
0b77d62 fixed it :-)

I have PQ sessions between iOS-WF and Whisperfish-WF now, and I think Android also worked! :tada:

libsignal-service/src/account_manager.rs
```diff
@@ -127,13 +145,36 @@ impl<Service: PushService> AccountManager<Service> {
             protocol_store
                 .save_pre_key(pre_key_id, &pre_key_record, None)
                 .await?;
+            // TODO: Shouldn't this also remove the previous pre-keys from storage?
```
gferon, line 148: this is actually a very good point, I'm actually not sure what happens then.

gferon (approved):
Awesome! This already looks good to me 😄

libsignal-service/src/account_manager.rs
```diff
@@ -163,11 +204,20 @@ impl<Service: PushService> AccountManager<Service> {
...
             signed_pre_key: signed_prekey_record.try_into()?,
             identity_key: *identity_key_pair.public_key(),
-            last_resort_key: if use_last_resort_key {
-                Some(PreKeyEntity {
-                    key_id: 0x7fffffff,
-                    public_key: "NDI=".into(),
-                })
+            pq_pre_keys: pq_pre_key_entities,
+            pq_last_resort_key: if use_last_resort_key {
+                log::warn!("Last resort Kyber key unimplemented");
```
gferon, line 209: do we maybe want to address this in the same PR, or do you think it's fine like this?

rubdos: It should be fine for us for now. Let's make an issue for it, because it actually requires revamping the pre-key code quite a bit, and we might want to do that in one go.

libsignal-service/src/account_manager.rs
rubdos: Will be the same revamp-pre-key issue.

PR #236, opened 2023-07-24
Remove prelude::{zkgroup,protocol} in favour of direct reexports

rubdos:
This removes a maintenance and development burden, and feels more logical and consistent to me. Dependees on libsignal-service can now drop the direct dep on libsignal-protocol/zkgroup/...

### DIFF
--- a/libsignal-service-actix/src/push_service.rs
+++ b/libsignal-service-actix/src/push_service.rs
@@ -551,10 +551,4 @@ mod tests {
             );
         }
     }
-
-    #[test]
-    fn serde_json_from_null() {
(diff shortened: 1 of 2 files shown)


shekohex (commented):
Looks good, but isn't that considered a breaking-change? I wound consider pumping up the version of the crate too, so other people that uses git to fetch the crate from github, they will get a cargo error.

rubdos:
> Looks good, but isn't that considered a breaking-change? I wound consider pumping up the version of the crate too, so other people that uses git to fetch the crate from github, they will get a cargo error.

We've never published on crates.io afaik.

PR #238, opened 2023-07-25
Registration fixes for 2023

rubdos:
The registration procedure has changed a few months ago, and the old API has been terminated. This is an attempt to expose the new registration API surface for Whisperfish. All the other clients I know are secondary-only, so this shouldn't affect anyone but Whisperfish.

### DIFF
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -58,6 +58,11 @@ pub const SENDER_CERTIFICATE_LEGACY_PATH: &str = "/v1/certificate/delivery";
 pub const SENDER_CERTIFICATE_PATH: &str =
     "/v1/certificate/delivery?includeUuid=true";
 
+pub const VERIFICATION_SESSION_PATH: &str = "/v1/verification/session";
+pub const VERIFICATION_CODE_PATH: &str    = "/v1/verification/session/%s/code";
+
+pub const REGISTRATION_PATH: &str    = "/v1/registration";
+
 pub const ATTACHMENT_DOWNLOAD_PATH: &str = "attachments/%d";
 
 pub const STICKER_MANIFEST_PATH: &str = "stickers/%s/manifest.proto";
@@ -225,6 +230,38 @@ pub struct WhoAmIResponse {
     pub pni: Uuid,
     #[serde(with = "serde_phone_number")]
     pub number: PhoneNumber,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RegistrationSessionMetadataResponse {
+    pub id: String,
+    #[serde(default)]
+    pub next_sms: Option<i32>,
+    #[serde(default)]
+    pub next_call: Option<i32>,
+    #[serde(default)]
+    pub next_verification_attempt: Option<i32>,
+    pub allowed_to_request_code: bool,
+    #[serde(default)]
+    pub requested_information: Vec<String>,
+    pub verified: bool,
+}
+
+impl RegistrationSessionMetadataResponse {
+    pub fn push_challenge_required(&self) -> bool {
+        // .contains() requires &String ...
+        self.requested_information
+            .iter()
+            .any(|x| x.as_str() == "pushChallenge")
+    }
+
+    pub fn captcha_required(&self) -> bool {
+        // .contains() requires &String ...
+        self.requested_information
+            .iter()
+            .any(|x| x.as_str() == "captcha")
+    }
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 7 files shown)


rubdos:
FWIW, Whisperfish diff: https://gitlab.com/whisperfish/whisperfish/-/merge_requests/481

MarcusGrass:
Don't know if this is related to this specific PR, but it seems that https://github.com/whisperfish/libsignal-service-rs/blob/main/libsignal-service/src/provisioning/manager.rs#L173 is now a raw string Ex `Got provisioning code: 'ac9e637e-16d1-50bd-b0aa-daf33b346cb0.1691236722275:7U2fte60zkioI3KLaIEZQT2skhYgFXLjIL2hgQqqLAQ='` Interestingly that worked yesterday

rubdos:
That would indeed be related, thanks for reporting. I have not been able to test this yet, because I don't have access to my second SIM.

MarcusGrass:
> That would indeed be related, thanks for reporting. I have not been able to test this yet, because I don't have access to my second SIM.

Yeah it's not easy with only one number, threw up my local fix as a separate PR here https://github.com/whisperfish/libsignal-service-rs/pull/240, it's nice in that it's just skipping some validation and shouldn't mess with any compatibility if someone were to unexpectedly have some old signal-compatible server they were testing against.

rubdos:
Thanks! I've cherry-picked your commit into this PR. Were you able to confirm a device with the other patches in here?

MarcusGrass:
> Thanks! I've cherry-picked your commit into this PR. Were you able to confirm a device with the other patches in here?

Nice, I can't test the registration flow at the moment because I also need another SIM, also, I'm using https://github.com/whisperfish/presage/ which breaks from the api-change here, would need to figure out how to rewire it first, do you know where the current sources for the java sdk is? All I can find are archived repos.

But the link-device flow works after that change on this branch!

I am getting this in the logs though: 
```
[2023-08-05T13:48:47Z ERROR libsignal_service::websocket] SignalWebSocket: Websocket error: end of web request stream; socket closing
[2023-08-05T13:48:51Z ERROR libsignal_service::websocket] SignalWebSocket: Websocket error: SignalWebSocket: end of application request stream; socket closing
```
It seems that the socket is closed from Signal's side https://github.com/whisperfish/libsignal-service-rs/blob/main/libsignal-service/src/websocket.rs#L231, then the stream starts returning `None`, although I don't think that's necessarily unexpected in this case since the connection is created and dropped for this request.

rubdos:
> do you know where the current sources for the java sdk is? All I can find are archived repos.

Everything is in Signal-Android these days. @gferon usually patches presage within a few days after I push things in here.



> It seems that the socket is closed from Signal's side https://github.com/whisperfish/libsignal-service-rs/blob/main/libsignal-service/src/websocket.rs#L231, then the stream starts returning `None`, although I don't think that's necessarily unexpected in this case since the connection is created and dropped for this request.

I think there's an issue about this. Because it doesn't necessarily harm the procedure, we didn't really bother at that point...

MarcusGrass:
> > do you know where the current sources for the java sdk is? All I can find are archived repos.
> 
> Everything is in Signal-Android these days. @gferon usually patches presage within a few days after I push things in here.
> 
Ah nice, thanks!

> > It seems that the socket is closed from Signal's side https://github.com/whisperfish/libsignal-service-rs/blob/main/libsignal-service/src/websocket.rs#L231, then the stream starts returning `None`, although I don't think that's necessarily unexpected in this case since the connection is created and dropped for this request.
> 
> I think there's an issue about this. Because it doesn't necessarily harm the procedure, we didn't really bother at that point...

Makes sense, it doesn't really seem problematic

gferon:
> That would indeed be related, thanks for reporting. I have not been able to test this yet, because I don't have access to my second SIM

FWIW, did you think of trying with your main SIM on staging servers? You could re-register infinitely (don't quote me on that) on the test servers.

rubdos:
Mea culpa, the confirm code was not part of the registration procedure at all. Any way useful to get it in here, it's part of registration anyway.

rubdos:
@direc85 confirms this works for Whisperfish.

rubdos:
The registration session thingy would be really cool in a nice RegistrationSession state machine structure, but I'm a bit too lazy to actually implement that now... Updating the examples.

rubdos:
@gferon this should be it. If you can throw some secondary eyes on it, feel free to hit merge!

libsignal-service/src/push_service.rs
```diff
@@ -227,6 +232,85 @@ pub struct WhoAmIResponse {
     pub number: PhoneNumber,
 }
 
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RegistrationSessionMetadataResponse {
```
gferon, line 237: Java :')

gferon (approved):
LGTM! I'll adapt `presage` immediately and check that it works as expected (although I read on the chat that somebody tried already.)

PR #240, opened 2023-08-05
Accept raw string confirmation code on confirm device

MarcusGrass:
Got a panic here today, seems to have been changed recently.

Removes u32 validation, since it seems signal isn't sticking to u32 codes. Besides, it's just converting to a u32 to then converting back to a string on the url. Backwards compatible.

### DIFF
--- a/libsignal-service/src/provisioning/manager.rs
+++ b/libsignal-service/src/provisioning/manager.rs
@@ -346,12 +346,11 @@ impl<P: PushService> LinkingManager<P> {
                         self.password.clone(),
                     );
 
-                    let provisioning_code = message
-                        .provisioning_code
-                        .ok_or(ProvisioningError::InvalidData {
-                            reason: "no provisioning confirmation code"
-                                .into(),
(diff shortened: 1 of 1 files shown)


rubdos:
Cherry-picked into #238

libsignal-service/src/provisioning/manager.rs
```diff
@@ -346,16 +346,16 @@ impl<P: PushService> LinkingManager<P> {
                         self.password.clone(),
                     );
 
+                    let provisioning_code = message
+                        .provisioning_code
+                        .ok_or(ProvisioningError::InvalidData {
+                            reason: "no provisioning confirmation code"
+                                .into(),
+                        })?;
```
rubdos, line 354:
```suggestion
                   let provisioning_code = message.provisioning_code.ok_or(
                       ProvisioningError::InvalidData {
                           reason: "no provisioning confirmation code".into(),
                       },
                   )?;
```

PR #241, opened 2023-08-08
Derive mcc/mnc inside of registration methods

gferon:
This was a sort of TODO in https://github.com/whisperfish/libsignal-service-rs/pull/238 but when adapting `presage` I felt like this would be better like suggested in the comment.

### DIFF
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -1186,7 +1186,7 @@ pub trait PushService: MaybeSend {
 
 trait PhoneNumberExt {
     fn mcc(&self) -> Option<&str>;
-    fn mnc(&self) -> Option<&str>; 
+    fn mnc(&self) -> Option<&str>;
 }
 
 impl PhoneNumberExt for PhoneNumber {
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -1020,10 +1020,8 @@ pub trait PushService: MaybeSend {
     // RegistrationSessionMetadataResponse createVerificationSession(@Nullable String pushToken, @Nullable String mcc, @Nullable String mnc)
     async fn create_verification_session<'a>(
         &mut self,
-        number: &'a str,
+        number: &'a PhoneNumber,
         push_token: Option<&'a str>,
-        mcc: Option<&'a str>,
-        mnc: Option<&'a str>,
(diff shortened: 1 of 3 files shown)


libsignal-service/src/push_service.rs
```diff
@@ -1058,10 +1058,9 @@ pub trait PushService: MaybeSend {
     // RegistrationSessionMetadataResponse patchVerificationSession(String sessionId, @Nullable String pushToken, @Nullable String mcc, @Nullable String mnc, @Nullable String captchaToken, @Nullable String pushChallengeToken)
     async fn patch_verification_session<'a>(
         &mut self,
+        number: &PhoneNumber,
```
rubdos, line 1061: This should be optional:

```suggestion
        number: Option<&PhoneNumber>,
```

libsignal-service/src/push_service.rs
```diff
@@ -1080,8 +1079,8 @@ pub trait PushService: MaybeSend {
             captcha,
             push_token_type: push_token.as_ref().map(|_| "fcm"),
             push_token,
-            mcc,
-            mnc,
+            mcc: number.mcc(),
+            mnc: number.mnc(),
```
rubdos, line 1083:
```suggestion
            mcc: number.and_then(PhoneNumberExt::mcc),
            mnc: number.and_then(PhoneNumberExt::mnc),
```

rubdos (commented):
PhoneNumber should be optional in the patch method

libsignal-service/src/push_service.rs
```diff
@@ -1184,3 +1183,18 @@ pub trait PushService: MaybeSend {
...
+
+impl PhoneNumberExt for PhoneNumber {
+    fn mcc(&self) -> Option<&str> {
+        self.carrier().and_then(|carrier| carrier.get(0..3))
+    }
+
+    fn mnc(&self) -> Option<&str> {
+        self.carrier().and_then(|carrier| carrier.get(3..))
+    }
+}
```
rubdos, line 1200: I was also thinking, maybe we want some tests on this code , and if it works, we can also put it in rust-phonenumber... Some tables here might help: https://en.wikipedia.org/wiki/Mobile_country_code

rubdos:
Closing in favour of https://github.com/whisperfish/rust-phonenumber/pull/59

PR #242, opened 2023-08-28
Stories APIs

rubdos:
This PR attempts to get all the APIs required for stories.

### DIFF
--- a/libsignal-service-actix/src/push_service.rs
+++ b/libsignal-service-actix/src/push_service.rs
@@ -43,11 +43,15 @@ impl AwcPushService {
         method: Method,
         endpoint: Endpoint,
         path: impl AsRef<str>,
+        additional_headers: &[(&str, &str)],
         credentials_override: HttpAuthOverride,
     ) -> Result<ClientRequest, ServiceError> {
         let url = self.cfg.base_url(endpoint).join(path.as_ref())?;
         log::debug!("HTTP request {} {}", method, url);
         let mut builder = self.client.request(method, url.as_str());
+        for &header in additional_headers {
+            builder = builder.insert_header(header);
+        }
         builder = match credentials_override {
             HttpAuthOverride::NoOverride => {
                 if let Some(credentials) = self.credentials.as_ref() {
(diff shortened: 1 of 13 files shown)


rubdos:
With the current API, clients will need to patch their message sender code for no meaningful reason:

```diff
diff --git a/whisperfish/src/worker/client.rs b/whisperfish/src/worker/client.rs
index daaa86c9..4b1849c0 100644
--- a/whisperfish/src/worker/client.rs
+++ b/whisperfish/src/worker/client.rs
@@ -330,7 +330,7 @@ impl ClientActor {
         let local_addr = self.local_addr.unwrap();
         let device_id = self.config.get_device_id();
         async move {
-            let u_ws = u_service.ws("/v1/websocket/", None, false).await?;
+            let u_ws = u_service.ws("/v1/websocket/", &[], None, false).await?;
             Ok(MessageSender::new(
                 ws,
                 u_ws,
```

rubdos:
This seems to correctly work with the Actix implementation.

libsignal-service-actix/src/push_service.rs
```diff
@@ -268,20 +271,24 @@ impl PushService for AwcPushService {
...
         S: Serialize,
     {
-        let mut response = self
-            .request(Method::PUT, endpoint, path, credentials_override)?
-            .send_json(&value)
-            .await
-            .map_err(|e| ServiceError::SendError {
+        let mut request =
+            self.request(Method::PUT, endpoint, path, credentials_override)?;
+        for &header in additional_headers {
```
gferon, line 284: how about adding an additional parameters for `additional_headers` to the `self.request()` method instead?

libsignal-service/src/push_service.rs
```diff
@@ -512,6 +512,8 @@ pub enum ServiceError {
     NotFoundError,
 }
 
+pub(crate) const NO_ADDITIONAL_HEADERS: &[(&str, &str)] = &[];
```
gferon, line 515: Why not, but that doesn't look very Rusty 😅 why not `Option<[]>` for the parameter so you can get rid of the constant?

libsignal-service/src/receiver.rs
```diff
@@ -44,10 +45,15 @@ impl<Service: PushService> MessageReceiver<Service> {
     pub async fn create_message_pipe(
         &mut self,
         credentials: ServiceCredentials,
+        allow_stories: bool,
     ) -> Result<MessagePipe, ServiceError> {
+        let headers = &[(
```
gferon, line 50: since `HeaderName::from_static` is `const fn` maybe we could instead store it in a constant (same for the value?)

gferon (approved):
My comments are mostly nits, since it already works are you'd expect. I just generally prefer stronger types 😄

libsignal-service/src/push_service.rs
rubdos: Yeh, the constant isn't very Rusty and doesn't really capture any intention. The only reason I used it, is because the method calls are clearer as to why there's a random `&[]` appearing practically everywhere.

I would vote against `Option<&[]>`, because then `None` and `&[]` would mean the same thing, and the implementations will have to `flat_map` them anyway.

Getting rid of the constant makes sense however.

libsignal-service-actix/src/push_service.rs
rubdos: ... because in my head somehow I read `client.request`, didn't notice it was a method we controlled and could change the signature of. Good catch!

libsignal-service/src/receiver.rs
gferon: Looks like `actix_web` re-exports `http` (https://docs.rs/actix-http/3.4.0/src/actix_http/header/map.rs.html#6) so it should be good to share the type. The `http` crate version also shouldn't be an issue at it's been _very_ stable.

rubdos: Honestly, if we're going there, I think we should just get rid of the split altogether and get Whisperfish to run on the Hyper backend. That would get rid of a lot more :bullettrain_front: :hankey: at the same time.

gferon: Fair point, feel free to merge as-is, but maybe having a single backend would also let us clean things up a little bit (not having an `async trait` for `PushService` and maybe a bit more magic as well). If you think it's the way to go, we can definitely do that.

rubdos: It's on my list of long-term things to get in Whisperfish. If we move to hyper some day, I'll certainly come back to that.

PR #243, opened 2023-09-04
Add server_guid to Metadata

rubdos:
This is unlike Signal Android, where the Envelope is also continously passed while processing the already decrypted content.  In Rust, we logically would *consume* the envelope to yield the Metadata and Content, so I prefer the Metadata to contain this information.

### DIFF
--- a/libsignal-service/src/cipher.rs
+++ b/libsignal-service/src/cipher.rs
@@ -103,6 +103,15 @@ where
             });
         };
 
+        let server_guid =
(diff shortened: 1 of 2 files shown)


gferon:
Can you add a small comment in the struct to explain what this field is?

rubdos:
> Can you add a small comment in the struct to explain what this field is?

I did just that (a few months ago, but I came across this again today)!

rubdos:
I'm just going to pull this in, it's a rather trivial change.

PR #246, opened 2023-09-22
Correct license specifiers

rubdos:
I'm messing around with `cargo deny`, and noticed that the licenses were unspecified in Cargo.toml for the implementation crates, and it was plain wrong in the main crate. License should be AGPL-3.0 for all.

We may want to start using `cargo deny` in CI here too, but I'll toy around in Whisperfish first.


gferon:
I use it at work for quite some time and it's nice!

rubdos:
> I use it at work for quite some time and it's nice!

Yeh it's quite cool. I have it on a scheduled pipeline and on every merge request now on Whisperfish, which should transitively also warn us about issues in here :-)

PR #247, opened 2023-09-23
Upgrade aes 0.7 -> 0.8

boxdot:
Now libsignal 0.32 is used, aes can be upgraded to the next version. In
particular, this removes duplicated dependencies.


libsignal-service/src/attachment_cipher.rs
```diff
@@ -30,12 +35,14 @@ pub fn encrypt_in_place(iv: [u8; 16], key: [u8; 64], plaintext: &mut Vec<u8>) {
     // Pad with zeroes for padding
     plaintext.extend(&[0u8; 16]);
 
-    let cipher = Cbc::<Aes256, Pkcs7>::new_from_slices(aes_half, &iv)
-        .expect("fixed length key material");
+    let cipher = Aes256CbcEnc::new(
+        aes_half.try_into().expect("fixed length key material"),
```
rubdos, line 39: I am very not much a fan of the `try_into()` call in these, these smell like non-const-time crypto.

rubdos (commented):
Could you comment on the `try_into()` that happens everywhere? Sounds like these should be using fixed-size arrays from the get-go, instead of dynamic conversions.

I would also suggest we merge #248 before this, then we know whether it is 1.61 compatible.

libsignal-service/src/attachment_cipher.rs
rubdos: Kind ping, @boxdot, now that #259 is in? :-)

boxdot: Thank you for pinging. I will finish it today!

boxdot: Replaced by `GenericArray::from_slice`

boxdot: Actually, it is possible just to use `slice.into()`, which also converts to `GenericArray` (it panics on wrong length).

rubdos: Maybe we can even change the `fn calculate_hmac256(` signature to return the `GenericArray`? Looks like it allocates to a `Vec<u8>` needlessly. Just looking at the code there, I'm not sure that it's actually the case.

boxdot: Done in 5f5577be0

PR #248, opened 2023-09-23
MSRV 1.61

rubdos:
Supersedes #149. Whisperfish will not be able to update libsignal-service for a while since libsignal-client 0.32, but we'll take that hit internally. You folks don't have to suffer anymore under 1.52 :-)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,7 +31,7 @@ jobs:
           args: --manifest-path ${{ matrix.project }}/Cargo.toml
 
   build:
-    name: Build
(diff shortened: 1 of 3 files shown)


gferon:
@rubdos I merged this but then realized that 1.61 currently fails because we require a lower minor version of `tempfile`. I suppose we need to check-in a lockfile?

rubdos:
Looks like the suffering at MSRV 1.61 has already started. Tokio has 1.63 as MSRV, prost claims 1.60, but in practice it's already 1.63 too. `tempfile` and a bunch of others (petgraph) are what kills the story transitively.

PR #249, opened 2023-10-02
Update protos to add support for EditMessage

gferon:
Unfortunately, it looks like the feature to send groups as attachments (for other people to join?) was deprecated. It looks like it was a feature for GV1 anyways. @rubdos I guess you can remove that from `whisperfish`?


rubdos:
> I guess you can remove that from `whisperfish`?

I don't think we have ever implemented group sync that way. Looks like a sync construction anyway. If it's an issue, I will remove it. Shouldn't be difficult.

Thanks!

PR #252, opened 2023-10-08
Ensure self-sending is unidentified

rubdos:
I previously thought this was incorrect, but I'm coming back from that. Untested. If someone else could test, that would be really neat :D

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -339,7 +339,7 @@ where
                     );
                 self.try_send_message(
                     self.local_address,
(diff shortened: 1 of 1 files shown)


gferon:
I think that might fix the bug with contacts 🎉

rubdos:
> I think that might fix the bug with contacts 🎉

Oops.

rubdos:
Ugh, I don't think this makes a lot of sense. You'd need your *own* US certificate to send to yourself, not the one of your conversation partner :'-)

PR #253, opened 2023-10-09
Propagate /v1/queue/empty to clients

gferon:
This is a much smaller API breakage than I expected it to be and will be a massive quality of life improvements for clients of `presage`. It can also be used to properly wait for contacts sync.


rubdos:
Android has refactored this API too, and there they return lists of Envelopes in batches. They also have a system in place where Envelopes can be acknowledged, instead of acking them in the stream itself.

That said, I think this design is perfectly fine.

PR #255, opened 2023-10-13
Add tracing

rubdos:


gferon:
Let's switch the calls to the log crate at the same time, then?

rubdos:
Yeh, there's a reason it's a draft ^^

Don't know when I'll continue though.

rubdos:
A cleaner trace for Envelopes would be beneficial:

```
2024-01-06 14:23:13 [INFO] decrypt; self=ServiceCipher { protocol_store: "...", csprng: "...", trust_root: "...", local_uuid: 9bad1xx, local_device_id: 1 } envelope=Envelope { r#type: Some(UnidentifiedSender), source_service_id: None, source_device: None, destination_service_id: Some("9bad"), timestamp: Some(170455xxx), content: Some([17, 10, 33, **LOADS MORE**, ...]), server_guid: Some("4d46a6"), server_timestamp: Some(17045509xxx), urgent: Some(false), story: Some(false), reporting_token: None }
```

libsignal-service-actix/src/push_service.rs
```diff
@@ -47,7 +47,7 @@ impl AwcPushService {
         credentials_override: HttpAuthOverride,
     ) -> Result<ClientRequest, ServiceError> {
         let url = self.cfg.base_url(endpoint).join(path.as_ref())?;
-        log::debug!("HTTP request {} {}", method, url);
+        tracing::debug!("HTTP request {} {}", method, url);
```
gferon, line 50: We can definitely merge this as is, but if you want to use the feature of `tracing` that adds support for structured logging, the syntax changes quite a bit.

```suggestion
        tracing::debug!(%url, %method, "HTTP request");
```

one can also rename fields in the logging statements, like so:

`tracing::debug!(http_url =% url, "HTTP request")`

You need to prefix the fields with `%` to use `fmt::Display` or `?` to use `fmt::Debug` when the type doesn't implement `tracing::Value`.

gferon:
> A cleaner trace for Envelopes would be beneficial:
> 
> ```
> 2024-01-06 14:23:13 [INFO] decrypt; self=ServiceCipher { protocol_store: "...", csprng: "...", trust_root: "...", local_uuid: 9bad1xx, local_device_id: 1 } envelope=Envelope { r#type: Some(UnidentifiedSender), source_service_id: None, source_device: None, destination_service_id: Some("9bad"), timestamp: Some(170455xxx), content: Some([17, 10, 33, **LOADS MORE**, ...]), server_guid: Some("4d46a6"), server_timestamp: Some(17045509xxx), urgent: Some(false), story: Some(false), reporting_token: None }
> ```

Most definitely, let's impl `tracing::Value` for it!

libsignal-service-actix/src/push_service.rs
rubdos: ... that's quite a QoL improvement and would remove the need for #129 / #53 !

rubdos:
> Most definitely, let's impl `tracing::Value` for it!

Turns out that is pretty difficult :'-)

One needs to implement `valuable::Valuable` (Value itself is Sealed), but Valuable is quite the involved visitor pattern.

rubdos:
I'm pulling spans over await points, and I shouldn't be doing that in `libsignal-service-actix/src/push_service.rs`. Needs fixing.

libsignal-service-actix/src/push_service.rs
rubdos: I changed quite some calls here. I suggest we clean up while we go, this is looking very pretty already on the console :)

rubdos:
> I'm pulling spans over await points, and I shouldn't be doing that in `libsignal-service-actix/src/push_service.rs`. Needs fixing.

I fixed most of these, let's get this pulled in! Looks rather pretty imo :)

PR #257, opened 2023-10-24
1.64 MSRV

rubdos:
It seems like we will have [Rust 1.64](https://github.com/sailfishos/rust/pull/19) on Sailfish OS soon™, so this makes our lives here a bit easier.

I'm still trying to get 1.72.1 through too, but I can't promise that yet.


rubdos:
Superceded by #259

PR #258, opened 2023-10-29
Update the libsignal-curve25519 lizard notice

rubdos:
@direc85 hit this undocumented change


gferon (approved):
Thought we had done this a while ago!

PR #259, opened 2023-11-05
Update some dependencies

direc85:
I'm updating Whisperfish to up-to-date dependencies, this results in dependencies with different versions. This PR attempts to lessen that duplication.

What's notable here, is that Sailfish OS should have Rust 1.72.1 soon*, so currently this PR pays no attention to MSRV.


direc85:
In the context of "Whisperfish is getting Rust 1.72 soon^tm", this PR is now in good enough shape to receive a proper code review. I believe @rubdos is also interested in this PR.

direc85:
I tried updating `aes` from 0.7 to 0.8 to (possible) further bring down the crate count, but I quickly hit a wall with futures, so if someone else wants to tackle that, feel free to do so!

direc85:
There's already #247 prepared. It looks like it applies cleanly on top of this PR and drops Whisperfish dependency count even further.

This also supersedes #257 (as some newer `actix` component requires Rust 1.70).

gferon:
LGTM, thanks a lot!

PR #260, opened 2023-11-07
fix: RUSTSEC-2023-0065 (part 2)

boxdot:
Fix security issue by upgrading tungstenite v0.19.0 -> v0.20.1.

Also specify MSRV in libsignal-service-hyper as 1.70.


rubdos:
Could you also change the MSRV in CI?

PR #263, opened 2023-11-19
Fix sending SyncMessage to self

gferon:
Fixes sending `SyncMessage` and to self.

- After looking at both Signal-Android and Signal-Desktop, we need to make sure any message sent to self is _never_ send unidentified
- Make sure we add padding to all `SyncMessage`
- Remove some redundant session loading before creating encrypted messages

### DIFF
--- a/libsignal-service/src/proto.rs
+++ b/libsignal-service/src/proto.rs
@@ -1,4 +1,6 @@
 #![allow(clippy::all)]
+
+use rand::{Rng, RngCore};
 include!(concat!(env!("OUT_DIR"), "/signalservice.rs"));
 include!(concat!(env!("OUT_DIR"), "/signal.rs"));
 
@@ -62,6 +64,20 @@ impl WebSocketResponseMessage {
                 message: Some("Unknown".to_string()),
                 ..Default::default()
             }
+        }
+    }
+}
+
+impl SyncMessage {
+    pub fn with_padding() -> Self {
+        let mut rng = rand::thread_rng();
+        let random_size = rng.gen_range(1..=512);
+        let mut padding: Vec<u8> = vec![0; random_size];
+        rng.fill_bytes(&mut padding);
+
+        Self {
+            padding: Some(padding),
+            ..Default::default()
         }
     }
 }
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -1,11 +1,11 @@
-use std::time::SystemTime;
+use std::{collections::HashSet, time::SystemTime};
 
 use chrono::prelude::*;
 use libsignal_protocol::{
     process_prekey_bundle, DeviceId, ProtocolStore, SenderCertificate,
     SenderKeyStore, SignalProtocolError,
 };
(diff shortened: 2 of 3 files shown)


libsignal-service/src/sender.rs
```diff
@@ -605,53 +617,35 @@ where
...
-            log::warn!("SessionStoreExt::get_sub_device_sessions should return unique device id's, and should not include DEFAULT_DEVICE_ID.");
-        }
+        // always send to the primary device no matter what
+        devices.insert(DEFAULT_DEVICE_ID.into());
 
-        // When sending to ourselves, don't include the local device
-        if myself {
-            devices.retain(|id| DeviceId::from(*id) != self.device_id);
+        // when sending a sync message, we always send it to all devices except the current one
+        if is_sync_message {
```
direc85, line 636: Should this be `myself && is_sync_message`? If my memory serves me right, `device_id` is a (small) integer so chances of device numbers colliding is significantly non-zero... Or, `is_sync_message` implies `myself = true` but if so, it would be clearer to have it explicitly stated here.

libsignal-service/src/sender.rs
```diff
@@ -605,53 +617,35 @@ where
...
-            )
-            .await?;
-            if self.protocol_store.load_session(&ppa).await?.is_some() {
+            debug!("sending message to device {}", device_id);
+            if device_id == DEFAULT_DEVICE_ID.into()
+                || self
+                    .protocol_store
+                    .load_session(&recipient.to_protocol_address(device_id))
+                    .await?
+                    .is_some()
```
direc85, line 647: This is a bit outside the scope of the PR, but why do we have this check in place? `create_encrypted_message` already checks and creates session if needed. Is that behavior avoided here for some specific reason? It would also bring down the repeated session loading, saving a few CPU cycles.

Edit: Okay, this is about creating the message content and not actually sending it, that's why. But doesn't auto-creating a message make sense anyway? If there is no session, the message won't be sent to that device, right..?

direc85 (approved):
Looks good to me, assuming some on-device testing has been made!

libsignal-service/src/sender.rs
gferon: This is what we were doing before but I rewrote the code for a reason that I ignore 😄

libsignal-service/src/sender.rs
gferon: I reworked the PR to not use `is_sync_message` since I finally understood how it's supposed to work.

libsignal-service/src/proto.rs
```diff
@@ -65,3 +67,17 @@ impl WebSocketResponseMessage {
         }
     }
 }
+
+impl SyncMessage {
+    pub fn with_padding() -> Self {
```
rubdos, line 72: @gferon Shouldn't we make those stronger types, such that it's impossible to transmit without padding?

gferon: Yes, that was a ninja trick for the time being.

PR #269, opened 2023-12-14
Enable keep-alive for provisioning socket

gferon:
Just found this while starting to look into linking errors.

Reference: https://github.com/signalapp/Signal-Desktop/blob/c6c072319993fefd9fcfab55ca77dc75263bc875/ts/textsecure/SocketManager.ts#L273


rubdos:
Are there no ws's anymore without KA? If so, don't forget to also amend actix :-)

gferon:
True dat, and yes they're all with keep alive.

PR #271, opened 2023-12-16
Fix linking as secondary device

gferon:
Extracted from https://github.com/whisperfish/presage/pull/220 with relevant refactoring and cleanups.


libsignal-service/src/provisioning/manager.rs
```diff
@@ -127,10 +84,15 @@ impl<P: PushService> LinkingManager<P> {
...
+        S: PreKeysStore,
+        R: rand::Rng + rand::CryptoRng,
+    >(
         &mut self,
+        aci_store: &mut S,
+        pni_store: &mut S,
         csprng: &mut R,
-        signaling_key: SignalingKey,
+        device_name: &str,
         mut tx: Sender<SecondaryDeviceProvisioning>,
```
boxdot, line 96: I would simplify this API. Just take a closure: `FnOnce(Url) -> Option<()>`.

boxdot: Also the `while let` loop below is misleading. It can be replaced by two calls to `next` and expecting url in the first call and message in the second call.

libsignal-service/src/provisioning/manager.rs
```diff
@@ -235,48 +171,145 @@ impl<P: PushService> LinkingManager<P> {
...
+                        device_id: u32,
+                    }
+
+                    let LinkResponse {
+                        aci,
+                        pni,
+                        device_id,
+                    } = self
+                        .push_service
+                        .put_json(
```
boxdot, line 286: I think this should become a method of the push service. Then the embedded structs can be also moved there.

libsignal-service/src/provisioning/manager.rs
```diff
@@ -235,48 +171,145 @@ impl<P: PushService> LinkingManager<P> {
...
+                        account_attributes: AccountAttributes,
+                        aci_signed_pre_key: SignedPreKey,
+                        pni_signed_pre_key: SignedPreKey,
+                        aci_pq_last_resort_pre_key: KyberPreKeyEntity,
+                        pni_pq_last_resort_pre_key: KyberPreKeyEntity,
+                    }
+
+                    #[derive(Debug, Serialize)]
+                    #[serde(rename_all = "camelCase")]
+                    struct AccountAttributes {
```
boxdot, line 224: I created this struct just for testing. There are already account attributes and capabilities. `DeviceCapabilities` are missing the `pni` field though.

libsignal-service/src/provisioning/manager.rs
```diff
@@ -235,48 +171,145 @@ impl<P: PushService> LinkingManager<P> {
...
+                            &[],
+                            HttpAuthOverride::Identified(HttpAuth {
+                                username: phone_number.to_string(),
+                                password: self.password.clone(),
+                            }),
+                            &request,
+                        )
+                        .await?;
+
+                    tx.send(
```
boxdot, line 298: No need for sending it back. Just return it from this async function.

libsignal-service/src/provisioning/manager.rs
```diff
@@ -40,82 +46,33 @@ pub struct ConfirmCodeResponse {
...
+    pub aci_public_key: PublicKey,
+    #[derivative(Debug = "ignore")]
+    pub pni_private_key: PrivateKey,
+    pub pni_public_key: PublicKey,
+    #[derivative(Debug = "ignore")]
+    pub profile_key: ProfileKey,
 }
 
 #[derive(Clone)]
 pub struct LinkingManager<P: PushService> {
```
boxdot, line 74: I think it makes sense to put the random generator also here inside.

libsignal-service/src/provisioning/manager.rs
boxdot: However I did not test if it links with these structs successfully. Maybe, they have some hard deprecated fields or similar. So, definitely need to test in case you decide to replace them.

gferon: Before we can cleanup, we need to check how we can fix/adapt registration as primary device, so I'd defer this to a later stage.

libsignal-service/src/provisioning/manager.rs
gferon: It was like this initially (a few years ago) and I remember users found it weird to have a function that waits for the end, while having to drive a rx with a `Url` simultaneously. I wouldn't change the API in this PR, but in another change.

libsignal-service/src/provisioning/manager.rs
gferon: Removed the loop in 097bb92

libsignal-service/src/provisioning/manager.rs
gferon: Done in 61113e

libsignal-service/src/provisioning/manager.rs
gferon: I actually would rather remove `LinkManager` since it's a struct that you never reuse, and I made the whole thing a free function called `link_device`.

libsignal-service/src/provisioning/manager.rs
boxdot: Sounds good to me.

boxdot (approved):

PR #275, opened 2024-01-04
Refactoring ACI/PNI storage types

rubdos:
Branch as I go along, with things that I disagree with :'-)

- Allow ACI/PNI store to be distinct types
- Make PreKeysStore async
- AccountManager can update prekey bundles for either PreKeyStore

One thing I disagree with, but don't necessarily want to fix here: `set_next_prekey_id` sounds like a great source of confusion. In RDBMS-based implementations, the `next_prekey_id`-family is expected to just query the database and return max+1. Setting a value doesn't make any sense, but the naming of the method implies it should need to be a method with a side effect.

### DIFF
--- a/libsignal-service-actix/examples/registering.rs
+++ b/libsignal-service-actix/examples/registering.rs
@@ -58,7 +58,7 @@ async fn main() -> Result<(), Error> {
     let mut push_service = AwcPushService::new(
         servers,
         Some(ServiceCredentials {
-            uuid: None,
+            aci: None,
             phonenumber: phonenumber.clone(),
             password,
             signaling_key: None,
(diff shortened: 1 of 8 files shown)


gferon (approved):
Let's ship it, I think I had a bit more quality of life improvements to add, but I can do it later! Thanks for this 😄

rubdos:
> Let's ship it, I think I had a bit more quality of life improvements to add, but I can do it later! Thanks for this 😄

FYI, this is 100% untested. I probably want to ship some more things, but https://gitlab.com/whisperfish/whisperfish/-/merge_requests/530 is on hold currently until after release :'-)

rubdos:
(This is more becoming the "let's just implement it damnit" merge request)

PR #277, opened 2024-01-08
Remove last remaining part of signaling key

rubdos:
Came across this while cleaning up Whisperfish at https://gitlab.com/whisperfish/whisperfish/-/merge_requests/534


rubdos:
Ugh, this is more difficult than anticipated. Signal Android still has the Signaling key in the AccountAttributes, but sets this to `null` all the time.

gferon:
> Ugh, this is more difficult than anticipated. Signal Android still has the Signaling key in the AccountAttributes, but sets this to `null` all the time.

Would adding a `Option<()>` field suffice? This is dumb but it'll work. We can even have a `serde_json::Value` set to `Null`.

PR #279, opened 2024-01-09
Cleanup some tracing

rubdos:
Cleaning up log messages that are too long while we test out the tracing support in Whisperfish


rubdos:
I'll have this in, then I can release a new Whisperfish :-)

Issue #54, opened 2020-12-21
New registration procedure?


rubdos:
I just got a Whisperfish log from a user, that contains the following (censored):

```
[2020-12-21T09:32:41Z DEBUG libsignal_service_actix::push_service] AwcPushService::put response: 
    ClientResponse HTTP/1.1 423 Locked
      headers:
        "date": "Mon, 21 Dec 2020 09:32:41 GMT"
        "content-type": "application/json"
        "x-signal-timestamp": "160854316xxxx"
        "connection": "keep-alive"
        "content-length": "181"
[2020-12-21T09:32:41Z TRACE libsignal_service_actix::push_service] Unhandled response with body: Ok(b"{\"timeRemaining\":NUMBER,\"backupCredentials\":{\"username\":\"4e3c26d8-UUID-XXX\",\"password\":\"4e3-UUID-XXX:NUMBER:SOMETHING-HEX\"}}")
[2020-12-21T09:32:41Z TRACE mio::poll] deregistering handle with poller
[2020-12-21T09:32:41Z ERROR harbour_whisperfish::worker::setup] Error in registration: Unexpected response: HTTP 423
```

I've never heard about `423 Locked`, do we know how to handle this?

PR #284, opened 2024-01-13
Propagate registration lock

direc85:
Add handler for `HTTP 423 Locked` to catch and pass registration lock error message during primary registration.

Contributes to #54


rubdos (approved):
Let's get this in so we can start testing it! :-)

PR #285, opened 2024-01-14
Implement PNI

### LINKED ISSUE #206: Add support for ACI/PNI

rubdos:
Fix #206, continuation of #275

TODO:
- [x] Update pre key bundle should take into account last-resort Kyber pre keys
- [x] Linking manager should use the account manager for pre key generation
- [x] Send the actual PNI distribution message
- [x] Use the supplied identity key pair in the refactored linking code
- [x] Consider refactoring `NewDeviceRegistration` to use `IdentityKey` instead of `PublicKey` or even straight `IdentityKeyPair`. Strong types ftw.

Maybe TODO:
- [ ] Phone number sharing is something that's stored on the profile. Add to the profile cipher.


rubdos:
@gferon: in the linking fix #271, you set some Kyber last resort keys:

```
        let aci_pq_last_resort_pre_key =
            generate_last_resort_kyber_key(aci_store, &aci_key_pair).await?;
        let pni_pq_last_resort_pre_key =
            generate_last_resort_kyber_key(pni_store, &pni_key_pair).await?;
```

The problem is that the storage does not realize this is a last resort key, and we should probably make sure to tag it as such.

Second, this pre key generation code is duplicate with AccountManager, we might want to clean that up.

gferon:
> @gferon: in the linking fix #271, you set some Kyber last resort keys:
> 
> ```
>         let aci_pq_last_resort_pre_key =
>             generate_last_resort_kyber_key(aci_store, &aci_key_pair).await?;
>         let pni_pq_last_resort_pre_key =
>             generate_last_resort_kyber_key(pni_store, &pni_key_pair).await?;
> ```
> 
> The problem is that the storage does not realize this is a last resort key, and we should probably make sure to tag it as such.
> 
> Second, this pre key generation code is duplicate with AccountManager, we might want to clean that up.

I remember adding a comment about this but didn't see some bad side effects 😅 and decided to go forward with the fix since everything was broken. Thanks for that!

rubdos:
[1c3d6dd](https://github.com/whisperfish/libsignal-service-rs/pull/285/commits/1c3d6dd9f3b31d927b2882300157ccff2b99aa78) refactors this, but is currently incorrect, as it relies on the `AccountManager` having knowledge of the newly provided identity key pairs.

rubdos:
Good for a first look. We'll test it thoroughly (as thoroughly as this complex thing can get tested...) with Whisperfish before getting this in.

libsignal-service/src/provisioning/mod.rs
```diff
@@ -211,18 +213,48 @@ pub async fn link_device<
...
+            aci_pq_last_resort_pre_key.expect("requested last resort key");
+        assert!(_aci_pre_keys.is_empty());
+        assert!(_aci_pq_pre_keys.is_empty());
+
+        let (
+            _pni_pre_keys,
+            pni_signed_pre_key,
+            _pni_pq_pre_keys,
+            pni_pq_last_resort_pre_key,
+        ) = crate::pre_keys::generate_pre_keys(
```
rubdos, line 245: This might be incorrect: we're using this to generate pre keys for other devices, so we *don't* have to store them ourselves, no? Needs a double check.

rubdos: Double check: we're generating them for ourselves, so this is fine!

rubdos:
Imagine. This works.

gferon:
Amazing, I'll integrate and test it in presage today!

rubdos:
> Amazing, I'll integrate and test it in presage today!

Quite a bit of work on the caller side, as usual, you can check out WF:
https://gitlab.com/whisperfish/whisperfish/-/merge_requests/550

direc85:
I've been using this with Whisperfish for a day or two now, and there haven't been anything weird to report about (after the initial shenanigans)!

gferon:
Just started the integration with presage, may I push small adjustment commits directly on this branch?

direc85:
I guess so, further fixes are always welcome. I'd still prefer the trial-and-error part happening in a fork or local check-out. Having said that, me and Ruben did exactly that a few days back :sweat_smile: Some final cleanup _may_ take place regardless.

rubdos:
> Just started the integration with presage, may I push small adjustment commits directly on this branch?

As long as you don't squash anything, I think that'd be fine.

rubdos:
@gferon I'm going to take the liberty to pull this in; @direc85 and I have been daily-driving this. It's not complete, but from here it's all incremental anyway.

gferon:
Go for it, I have a bunch of changes that I'll try to commit in separate changes, since they are mostly quality of life improvements and I went a little bit extra with my changes 😂

PR #286, opened 2024-01-14
Send SyncMessages to self unidentified when possible

direc85:
This changes the approach of sending `SyncMessage`s quite a lot, yet it shouldn't be _too_ invasive. This has received a brief test run in a device and it improves the situation quite a lot, if not fixes the core issue as-is! (knocks on wood)

- Add self-sync-message results to group and dm message sender results, and so...
- Return a `Vec<...>` from `send_message()`
- Accept the self recipient unidentified access when sending DMs
- Never send any "oridinary messages" to self, but only `SyncMessages`

I'll continue running this on my device for a while to see if there are any ill behaviors after this, and polish and fix as needed before un-drafting this.

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -315,8 +315,8 @@ where
             _ => false,
         };
 
-        // don't send anything to self nor session enders to others as sealed sender
-        if recipient == &self.local_address || end_session {
+        // don't send session enders as sealed sender
+        if end_session {
             unidentified_access.take();
         }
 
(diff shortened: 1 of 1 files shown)


direc85:
Note that this also changes the API a bit, so let's try to be in sync with this :)

direc85:
Superseded by #287 (and [Whisperfish#594](https://gitlab.com/whisperfish/whisperfish/-/merge_requests/549))

It turns out that sending SyncMessages nor Messages to self recipient identified works, so I'm closing this.

PR #287, opened 2024-01-15
Attempt to reestablish a session when it's found to be corrupt.

rubdos:
@direc85 and me found some corrupt sessions in the database. Since https://github.com/whisperfish/libsignal-client/commit/601454d20, libsignal-protocol forwards corruption as `SessionNotFound` errors, which makes me believe we're allowed to recreate said session. This patch implements exactly that.

This deviates from the behaviour at the Android side, as far as I can tell, so needs caution. I tested this in the Whisperfishers group, seems to behave as intended.

FWIW, I blame said session corruption on going back-and-forth between the Kyber and pre-Kyber code a few times. I recall doing that and not turning off my network by accident.

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -663,6 +663,9 @@ where
 
         for device_id in devices {
             trace!("sending message to device {}", device_id);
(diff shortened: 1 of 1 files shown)


gferon:
Nice! I'll give it a look tomorrow morning if that's ok.

PR #289, opened 2024-01-28
Fix base64

### LINKED ISSUE #288: JSON Invalid Padding
Sometimes the data returned by Signal is in base64-format and not properly padded. Those errors were ignored by the `base64`-crate with version `0.13`, but the dependency update in #280 updated `base64` to `0.21` where this error is now thrown.

This is for me constantly reproducible with one group, where a missmatched device always returns invalid padded code.

To fix this, one probably has to replace every use of `BASE64_STANDARD` with `BASE64_STANDARD_NO_PAD`.

Related upstream error: <https://gitlab.com/schmiddi-on-mobile/flare/-/issues/157>. This was first noted by @boxdot in https://github.com/whisperfish/presage/pull/230#issuecomment-1899040247.

Schmiddiii:
Fixes #288.

Note that `BASE64_STANDARD_NO_PAD` does not work as this requires base64 to not have padding, which is also not the case for Signal. This therefore needs a custom engine.

### DIFF
--- a/libsignal-service/src/utils.rs
+++ b/libsignal-service/src/utils.rs
@@ -1,4 +1,18 @@
+// Signal sometimes adds padding, sometimes it does not.
+// This requires a custom decoding engine.
+// This engine is as general as possible.
+pub const BASE64_RELAXED: base64::engine::GeneralPurpose =
+    base64::engine::GeneralPurpose::new(
+        &base64::alphabet::STANDARD,
+        base64::engine::GeneralPurposeConfig::new()
+            .with_encode_padding(true)
+            .with_decode_padding_mode(
+                base64::engine::DecodePaddingMode::Indifferent,
+            ),
+    );
+
 pub mod serde_base64 {
+    use super::BASE64_RELAXED;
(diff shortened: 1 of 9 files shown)


rubdos:
Seems like you miss some imports in `account_manager.rs`.

libsignal-service/src/utils.rs
```diff
@@ -1,4 +1,18 @@
+// Signal sometimes adds padding, sometimes it does not.
+// This requires a custom decoding engine.
+// This engine is as general as possible.
+pub const BASE64: base64::engine::GeneralPurpose =
```
rubdos, line 4: I would give this a slightly more descriptive name, like this:

```suggestion
pub const BASE64_RELAXED: base64::engine::GeneralPurpose =
```

If you want, I'll run Rust analyzer over the source and get that committed, if you don't have a rename command around.

Other name suggestions are welcome too.

Schmiddiii: Renamed as desired.

PR #290, opened 2024-02-13
Publish profile avatar URL

Schmiddiii:
This is to allow presage to fetch profile avatars.

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -69,6 +69,7 @@ pub struct Profile {
     pub name: Option<ProfileName<String>>,
     pub about: Option<String>,
     pub about_emoji: Option<String>,
+    pub avatar: Option<String>,
 }
 
 const PRE_KEY_MINIMUM: u32 = 10;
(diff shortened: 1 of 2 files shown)


Schmiddiii:
(CI failure is most likely not my fault, something nightly-related)

Schmiddiii:
> Hmm, how comes Whisperfish could show profiles? Am I using some API that I should not? :D

You are directly using `SignalServiceProfile`, which I did not want to use with presage as this seems pretty low-level and also contains encrypted data.

> (CI failure is most likely not my fault, something nightly-related)

This seems to be https://github.com/dalek-cryptography/curve25519-dalek/issues/618. Not sure what we can do about that as `libsignal` depends on its own version of `curve25519-dalek` and this does not yet have version `4.1.2` available, which would fix this.

rubdos:
> > (CI failure is most likely not my fault, something nightly-related)
> 
> This seems to be [dalek-cryptography/curve25519-dalek#618](https://github.com/dalek-cryptography/curve25519-dalek/issues/618). Not sure what we can do about that as `libsignal` depends on its own version of `curve25519-dalek` and this does not yet have version `4.1.2` available, which would fix this.

Looks like that will be fixed automagically anyway, so this sounds mergable to me.

PR #291, opened 2024-02-21
Fix panic for fetching group avatar

Schmiddiii:
Tracing assumed the path was always at least 4 characters long. This would panic with the empty path which can be the value of a group avatar path if no avatar is set for the group.

While this panic is easy to prevent in presage (which I also did [here](https://github.com/whisperfish/presage/pull/234/commits/1852f52c5b076bdb0a2dc0e6972e5096e7ab6fa6#diff-99551c60a775f81baa1c0f5ba314ab329a8b187a979ece33826b3f98a02c3fa3R502)), this panic should still be removed from libsignal-service-rs.

(CI failure is again in nightly and most likely not my fault)

### DIFF
--- a/libsignal-service/src/groups_v2/manager.rs
+++ b/libsignal-service/src/groups_v2/manager.rs
@@ -267,7 +267,7 @@ impl<S: PushService, C: CredentialsCache> GroupsManager<S, C> {
 
     #[tracing::instrument(
         skip(self, group_secret_params),
-        fields(path = %path[..4]),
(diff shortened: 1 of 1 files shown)


shekohex (approved):

rubdos:
> This would panic with the empty path which can be the value of a group avatar path if no avatar is set for the group.

I don't think you should call this method when you know the avatar is not set, but I agree it shouldn't panic in that case.

PR #293, opened 2024-03-01
Update libsignal-protocol, zkgroup and curve25519-dalek

direc85:
As  @boxdot  said in `#whisperfish` Matrix channel:

> boxdot: We are quite some versions behind libsignal-protocol: using v0.32.0, last released 0.40.1
> boxdot: WDYT about upgrading?
> boxdot: It actually builds out of the box

Update libsignal-protocol and zkgroup to v0.40.1 and curve25519-dalek to signal-curve25519-4.1.1.

Updating tests was also necessary.

Tested on-device with Whisperfish and it seems to works fine for a few sent and received messages, both in a DM and in a group.


gferon:
Neat, thanks! I'll test ASAP.

gferon:
Works fine in my tests, merging! 🚀

rubdos:
> Updating tests was also necessary.

I don't see any updated tests?

direc85:
> I don't see any updated tests?

My bad. The code compiled, but `push_service::tests::create_clients` test failed. I fixed the pointed piece of code (`configuration.rs`) with updating the offending parts from the sources pointed in the comments, but didn't realize that what I fixed _wasn't_ a test.

PR #297, opened 2024-03-18
Fix Groups

### LINKED ISSUE #296: Groups Fail with Newly Linked Devices
Currently, linking Flare (and therefore presage and libsignal-service-rs) as a new device does not work with groups. The following error is printed:

```
failed to receive auth credentials with PNI: ZkGroupVerificationFailure
```

The source of this error is [here](https://github.com/whisperfish/libsignal-service-rs/blob/b0fd817d616e409d32cc7d317edb423b1cbb1e4d/libsignal-service/src/groups_v2/manager.rs#L209).

@rubdos mentioned in the Whisperfish signal chat that this may be that the `/v1/certificate/auth/group`-API potentially changed, looking at the [Signal-Desktop code](https://github.com/signalapp/Signal-Desktop/blob/f77f9165e1100c3cae760a77a56e8a76e7ecf9bc/ts/textsecure/WebAPI.ts#L3137) there indeed seems to be a new parameter `pniAsServiceId=true`. I would guess that the call without that parameter got removed which causes the failure. Maybe adding it fixes it, it may also require replacing `receive_auth_credential_with_pni_as_aci` with `receive_auth_credential_with_pni_as_service_id` [here](https://github.com/whisperfish/libsignal-service-rs/blob/b0fd817d616e409d32cc7d317edb423b1cbb1e4d/libsignal-service/src/groups_v2/manager.rs#L201). I did not test my guess yet though.

Schmiddiii:
Fixes #296.

### DIFF
--- a/libsignal-service/src/groups_v2/manager.rs
+++ b/libsignal-service/src/groups_v2/manager.rs
@@ -161,7 +161,7 @@ impl<S: PushService, C: CredentialsCache> GroupsManager<S, C> {
             auth_credential_response
         } else {
             let path =
-            format!("/v1/certificate/auth/group?redemptionStartSeconds={}&redemptionEndSeconds={}", today, today_plus_7_days);
+            format!("/v1/certificate/auth/group?redemptionStartSeconds={}&redemptionEndSeconds={}&pniAsServiceId=true", today, today_plus_7_days);
 
             let credentials_response: CredentialResponse = self
                 .push_service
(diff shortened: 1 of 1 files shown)


gferon:
Great find!

rubdos:
Sounds like we should also expose the false-equivalent.

Schmiddiii:
Both in recent versions of Signal-Desktop and Signal-Android, this value is hardcoded to `true`. So not sure we should expose a false-equivalent.

rubdos:
> Both in recent versions of Signal-Desktop and Signal-Android, this value is hardcoded to `true`. So not sure we should expose a false-equivalent.

Awesome, thanks for sharing that. Not even the `/v1/certificate/auth/group?redemptionStartSeconds={}&redemptionEndSeconds={}` without the additional `pniAsServiceId`?

rubdos:
> Awesome, thanks for sharing that. Not even the `/v1/certificate/auth/group?redemptionStartSeconds={}&redemptionEndSeconds={}` without the additional `pniAsServiceId`?

Nope! So we don't have to expose it for now. Cool.

PR #298, opened 2024-04-02
Various cleanups after merging ACI/PNI

gferon:
Various small changes I did as part of https://github.com/whisperfish/presage/pull/245:

- Use `enum` variants in a bunch of errors instead of static strings
- Rename `VerifyAccountResponse::uuid` to `VerifyAccountResponse::aci` for completion (but use a `serde` alias)
- Other _kleinigkeiten_
- Use `IdentityKey` in some places where `PublicKey` was used before (and change the relevant `serde` modules)


libsignal-service/src/session_store.rs
```diff
@@ -50,44 +50,36 @@ pub trait SessionStoreExt: SessionStore {
 
     // #[async_trait] sadly doesn't work here,
     // because calls to libsignal_protocol are unsend.
-    fn compute_safety_number<'s>(
+    async fn compute_safety_number<'s>(
```
rubdos, line 53: I take it that this comment is superseded, then?

gferon: _maybe_ I need to check that it doesn't only work with `hyper` but IIRC the restriction in `libsignal` was removed.

libsignal-service/src/pre_keys.rs
```diff
@@ -180,7 +190,7 @@ pub(crate) const PRE_KEY_MINIMUM: u32 = 10;
 pub(crate) const PRE_KEY_BATCH_SIZE: u32 = 100;
 pub(crate) const PRE_KEY_MEDIUM_MAX_VALUE: u32 = 0xFFFFFF;
 
-pub(crate) async fn replenish_pre_keys<
+pub async fn replenish_pre_keys<
     R: rand::Rng + rand::CryptoRng,
```
rubdos, line 194: I'm not necessarily convinced we want this exposed outside of the crate.

rubdos (commented):
Looks rather good to me! :)

libsignal-service/src/pre_keys.rs
gferon: I'll check, might be a remnant, but it might also be me being lazy with the current state of the linking code between this and `presage`.

gferon: Restored!

rubdos:
Cool, feel free to merge this! :-)

rubdos (approved):
:+1:

PR #299, opened 2024-05-17
Better handle PNI UUID

direc85:
Try to handle PNI UUID better where ACI UUID is expected.

### DIFF
--- a/libsignal-service/examples/storage.rs
+++ b/libsignal-service/examples/storage.rs
@@ -9,6 +9,12 @@ use libsignal_service::protocol::{
 #[derive(Default)]
 pub struct ExampleStore {}
 
+impl Default for ExampleStore {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl ExampleStore {
     pub fn new() -> Self {
         Self::default()
--- a/libsignal-service/examples/storage.rs
+++ b/libsignal-service/examples/storage.rs
@@ -9,12 +9,6 @@ use libsignal_service::protocol::{
 #[derive(Default)]
 pub struct ExampleStore {}
 
-impl Default for ExampleStore {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl ExampleStore {
     pub fn new() -> Self {
         Self::default()
--- a/libsignal-service/src/profile_service.rs
+++ b/libsignal-service/src/profile_service.rs
@@ -21,9 +21,10 @@ impl ProfileService {
     ) -> Result<SignalServiceProfile, ServiceError> {
         let endpoint = match profile_key {
             Some(key) => {
-                let version = bincode::serialize(
-                    &key.get_profile_key_version(address.aci().unwrap()),
-                )?;
+                let version =
+                    bincode::serialize(&key.get_profile_key_version(
+                        address.aci().expect("profile by ACI ProtocolAddress"),
+                    ))?;
                 let version = std::str::from_utf8(&version)
                     .expect("hex encoded profile key version");
                 format!("/v1/profile/{}/{}", address.uuid, version)
--- a/libsignal-service/src/profile_service.rs
+++ b/libsignal-service/src/profile_service.rs
@@ -22,7 +22,7 @@ impl ProfileService {
         let endpoint = match profile_key {
             Some(key) => {
                 let version = bincode::serialize(
-                    &key.get_profile_key_version(address.aci()),
+                    &key.get_profile_key_version(address.aci().unwrap()),
                 )?;
                 let version = std::str::from_utf8(&version)
                     .expect("hex encoded profile key version");
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -75,7 +75,7 @@ pub const STICKER_PATH: &str = "stickers/%s/full/%d";
 pub const KEEPALIVE_TIMEOUT_SECONDS: Duration = Duration::from_secs(55);
 pub const DEFAULT_DEVICE_ID: u32 = 1;
 
-#[derive(Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize, Hash)]
 pub enum ServiceIdType {
     /// Account Identity (ACI)
     ///
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -75,7 +75,7 @@ pub const STICKER_PATH: &str = "stickers/%s/full/%d";
 pub const KEEPALIVE_TIMEOUT_SECONDS: Duration = Duration::from_secs(55);
 pub const DEFAULT_DEVICE_ID: u32 = 1;
 
-#[derive(Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize, Hash)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum ServiceIdType {
     /// Account Identity (ACI)
     ///
@@ -811,7 +811,7 @@ pub trait PushService: MaybeSend {
         &mut self,
         messages: OutgoingPushMessages,
     ) -> Result<SendMessageResponse, ServiceError> {
-        let path = format!("/v1/messages/{}", messages.recipient.uuid);
+        let path = format!("/v1/messages/{}", messages.destination);
         self.put_json(
             Endpoint::Service,
             &path,
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -75,7 +75,7 @@ pub const STICKER_PATH: &str = "stickers/%s/full/%d";
 pub const KEEPALIVE_TIMEOUT_SECONDS: Duration = Duration::from_secs(55);
 pub const DEFAULT_DEVICE_ID: u32 = 1;
 
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize)]
 pub enum ServiceIdType {
     /// Account Identity (ACI)
     ///
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -903,7 +903,7 @@ pub trait PushService: MaybeSend {
     ) -> Result<SignalServiceProfile, ServiceError> {
         let endpoint = if let Some(key) = profile_key {
             let version = bincode::serialize(
-                &key.get_profile_key_version(address.aci()),
+                &key.get_profile_key_version(address.aci().unwrap()),
             )?;
             let version = std::str::from_utf8(&version)
                 .expect("hex encoded profile key version");
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -902,9 +902,9 @@ pub trait PushService: MaybeSend {
         profile_key: Option<zkgroup::profiles::ProfileKey>,
     ) -> Result<SignalServiceProfile, ServiceError> {
         let endpoint = if let Some(key) = profile_key {
-            let version = bincode::serialize(
-                &key.get_profile_key_version(address.aci().unwrap()),
-            )?;
+            let version = bincode::serialize(&key.get_profile_key_version(
+                address.aci().expect("profile by ACI ProtocolAddress"),
+            ))?;
             let version = std::str::from_utf8(&version)
                 .expect("hex encoded profile key version");
             format!("/v1/profile/{}/{}", address.uuid, version)
--- a/libsignal-service/src/service_address.rs
+++ b/libsignal-service/src/service_address.rs
@@ -4,6 +4,8 @@ use libsignal_protocol::{DeviceId, ProtocolAddress};
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
+pub use crate::push_service::ServiceIdType;
+
 #[derive(thiserror::Error, Debug, Clone)]
 pub enum ParseServiceAddressError {
     #[error("Supplied UUID could not be parsed")]
@@ -16,6 +18,7 @@ pub enum ParseServiceAddressError {
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct ServiceAddress {
     pub uuid: Uuid,
+    pub identity: ServiceIdType,
 }
 
 impl ServiceAddress {
@@ -37,7 +40,10 @@ impl ServiceAddress {
 
 impl From<Uuid> for ServiceAddress {
     fn from(uuid: Uuid) -> Self {
-        Self { uuid }
+        Self {
+            uuid,
+            identity: ServiceIdType::AccountIdentity,
+        }
     }
 }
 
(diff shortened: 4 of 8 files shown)


shekohex (approved):

libsignal-service/src/service_address.rs
```diff
@@ -37,27 +40,55 @@ impl ServiceAddress {
...
+            Ok(ServiceAddress {
+                uuid: Uuid::from_slice(value.strip_prefix(b"PNI:").unwrap())?,
+                identity: ServiceIdType::PhoneNumberIdentity,
+            })
+        } else {
+            Ok(ServiceAddress {
+                uuid: Uuid::from_slice(value)?,
+                identity: ServiceIdType::AccountIdentity,
+            })
+        }
```
rubdos, line 82:
```suggestion
        if let Some(pni) = value.strip_prefix("bPNI:") {
            Ok(ServiceAddress {
                uuid: Uuid::from_slice(pni)?,
                identity: ServiceIdType::PhoneNumberIdentity,
            })
        } else {
            Ok(ServiceAddress {
                uuid: Uuid::from_slice(value)?,
                identity: ServiceIdType::AccountIdentity,
            })
        }
```

Fewer unwraps :)

rubdos (commented):
I'd love to ensure somehow that we can't get the UUID without having checked whether we're at ACI or PNI. Getting this type safe would be ideal.

libsignal-service/examples/storage.rs
```diff
@@ -8,6 +8,12 @@ use libsignal_service::protocol::{
...
 
+impl Default for ExampleStore {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl ExampleStore {
     pub fn new() -> Self {
         Self {}
```
rubdos, line 19:
```suggestion
#[derive(Default)]
pub struct ExampleStore {}

impl ExampleStore {
    pub fn new() -> Self {
        Self::default()
```

libsignal-service/src/push_service.rs
```diff
@@ -72,7 +72,7 @@ pub const STICKER_PATH: &str = "stickers/%s/full/%d";
 pub const KEEPALIVE_TIMEOUT_SECONDS: Duration = Duration::from_secs(55);
 pub const DEFAULT_DEVICE_ID: u32 = 1;
 
-#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize)]
```
rubdos, line 75: It's not necessarily immediately clear to me how this enum gets serialized; shouldn't we just be using `to_string()` on instances?

libsignal-service/src/service_address.rs
```diff
@@ -33,31 +36,93 @@ impl ServiceAddress {
     pub fn pni(&self) -> libsignal_protocol::Pni {
         libsignal_protocol::Pni::from_uuid_bytes(self.uuid.into_bytes())
     }
```
rubdos, line 38: Since we now have strong types in this ServiceAddress struct, should we fail the `aci`/`pni` methods when the type is incorrect? We could also rename the methods to `unwrap_aci` and `unwrap_pni`. Currently, we allow "converting" an ACI UUID into a PNI and vice versa, I think.

libsignal-service/src/service_address.rs
```diff
@@ -33,31 +36,93 @@ impl ServiceAddress {
...
+                    identity: ServiceIdType::AccountIdentity,
+                }),
+                Err(e) => {
+                    tracing::error!("Unknown identity format: {}", value);
+                    Err(ParseServiceAddressError::InvalidUuid(e))
+                },
+            }
+        }
     }
 }
```
rubdos, line 82: If we can convert from a `ProtocolAddress`, then our `to_protocol_address` should probably also format a PNI as `PNI:{uuid}`? I'm not sure how this maps on Signal-Android, and how SA/SD/iS handle the ServiceAddress, but I'd vote we keep it as close as possible.

libsignal-service/src/service_address.rs
```diff
@@ -33,31 +36,93 @@ impl ServiceAddress {
...
+        } else {
+            Ok(ServiceAddress {
+                uuid: Uuid::from_slice(value)?,
+                identity: ServiceIdType::AccountIdentity,
+            })
+        }
     }
 }
 
 impl TryFrom<Option<&str>> for ServiceAddress {
```
rubdos, line 120: We should probably just deprecate or remove those two...

rubdos (commented):
Getting this as close in line with upstream implementations is really mandatory, but I'm not sure how they currently do this. Of course, we also want it as Rusty as possible...

libsignal-service/src/push_service.rs
direc85: I'll have to double check why that was necessary. I added those when I tried to make this more Diesel-friendly, but `Display` also exists currently.

libsignal-service/src/service_address.rs
direc85: We are currently having a lot of ACI/PNI UUID representations, perhaps we could trim some of them down at some point...

Yes, we should fail the methods if the type is incorrect. Renaming the methods sounds good as well.

libsignal-service/src/service_address.rs
direc85: My approach is based on Signal sending UUID's in the format of `Some("{uuid}")` and `Some("PNI:{uuid}")` so that's why I try to maintain that representation too. One assumption is that `ProtocolAddress` can contain the `PNI:` prefix too, but that's indeed unverified w.r.t. e.g. Signal Android code base.

I'll check the `to_protocol_address` method, sounds like I've missed that one.

libsignal-service/src/service_address.rs
direc85: IIRC `try_from` was used in some places and it resulted in quite a nice and Rust-y code. I'll see what happens when I try to remove those.

libsignal-service/src/service_address.rs
rubdos: We could also keep `aci` and make it return an Option instead.

libsignal-service/src/service_address.rs
```diff
@@ -138,3 +138,9 @@ impl TryFrom<Option<&[u8]>> for ServiceAddress {
         }
     }
 }
+
+impl Hash for ServiceAddress {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.to_service_id().hash(state);
+    }
+}
```
rubdos, line 146: Shoudl we just derive Hash on ServiceAddress?

libsignal-service/examples/storage.rs
direc85: Fixed, thanks!

libsignal-service/src/push_service.rs
direc85: It would be this (cleaned up):

```
   Compiling libsignal-service v0.1.0 (/home/matti/SFOS/libsignal-service-rs/libsignal-service)
error[E0277]: the trait bound `ServiceIdType: Serialize` is not satisfied
    --> /home/matti/SFOS/libsignal-service-rs/libsignal-service/src/service_address.rs:18:45
     |
18   | #[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize)]
     |                                             ^^^^^^^^^ the trait `Serialize` is not implemented for `ServiceIdType`
...
21   |     pub identity: ServiceIdType,
     |     --- required by a bound introduced by this call
     |
note: required by a bound in `account_manager::_::_serde::ser::SerializeStruct::serialize_field`
    --> /home/matti/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.197/src/ser/mod.rs:1865:12
     |
1859 |     fn serialize_field<T: ?Sized>(
     |        --------------- required by a bound in this associated function
...
1865 |         T: Serialize;
     |            ^^^^^^^^^ required by this bound in `SerializeStruct::serialize_field`

error[E0277]: the trait bound `ServiceIdType: Deserialize<'_>` is not satisfied
    --> /home/matti/SFOS/libsignal-service-rs/libsignal-service/src/service_address.rs:21:19
     |
21   |     pub identity: ServiceIdType,
     |                   ^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `ServiceIdType`
     |
note: required by a bound in `next_element`
    --> /home/matti/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.197/src/de/mod.rs:1726:12
     |
1724 |     fn next_element<T>(&mut self) -> Result<Option<T>, Self::Error>
     |        ------------ required by a bound in this associated function
1725 |     where
1726 |         T: Deserialize<'de>,
     |            ^^^^^^^^^^^^^^^^ required by this bound in `SeqAccess::next_element`

error[E0277]: the trait bound `ServiceIdType: Deserialize<'_>` is not satisfied
    --> /home/matti/SFOS/libsignal-service-rs/libsignal-service/src/service_address.rs:21:19
     |
21   |     pub identity: ServiceIdType,
     |                   ^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `ServiceIdType`
     |
note: required by a bound in `next_value`
    --> /home/matti/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.197/src/de/mod.rs:1865:12
     |
1863 |     fn next_value<V>(&mut self) -> Result<V, Self::Error>
     |        ---------- required by a bound in this associated function
1864 |     where
1865 |         V: Deserialize<'de>,
     |            ^^^^^^^^^^^^^^^^ required by this bound in `MapAccess::next_value`

error[E0277]: the trait bound `ServiceIdType: Deserialize<'_>` is not satisfied
  --> /home/matti/SFOS/libsignal-service-rs/libsignal-service/src/service_address.rs:21:5
   |
21 |     pub identity: ServiceIdType,
   |     ^^^ the trait `Deserialize<'_>` is not implemented for `ServiceIdType`
   |
note: required by a bound in `account_manager::_::_serde::__private::de::missing_field`
  --> /home/matti/.cargo/registry/src/index.crates.io-6f17d22bba15001f/serde-1.0.197/src/private/de.rs:25:8
   |
23 | pub fn missing_field<'de, V, E>(field: &'static str) -> Result<V, E>
   |        ------------- required by a bound in this function
24 | where
25 |     V: Deserialize<'de>,
   |        ^^^^^^^^^^^^^^^^ required by this bound in `missing_field`
```
So, I guess since it goes down to Diesel, (`SeqAccess::next_element`) they have to be kept in?

rubdos: I don't think Diesel uses Serialize. I did some digging.

`ServiceAddress` now has the `ServiceIdType` as field. `ServiceAddress` gets serialized when calling `PushService::send_messages`. However, the Signal Server *ignores* unknown fields, and it never reads the `pub recipient: ServiceAddress,` which we currently have, nor the `@JsonProperty private String destination;` which Signal Android currently uses as a replacement.

So to mimic the Android client better, I would suggest:

```diff
diff --git a/libsignal-service/src/push_service.rs b/libsignal-service/src/push_service.rs
index 9cd10d12a..a2383c843 100644
--- a/libsignal-service/src/push_service.rs
+++ b/libsignal-service/src/push_service.rs
@@ -75,7 +75,7 @@ pub const STICKER_PATH: &str = "stickers/%s/full/%d";
 pub const KEEPALIVE_TIMEOUT_SECONDS: Duration = Duration::from_secs(55);
 pub const DEFAULT_DEVICE_ID: u32 = 1;
 
-#[derive(Debug, Clone, Copy, Eq, PartialEq, Serialize, Deserialize, Hash)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum ServiceIdType {
     /// Account Identity (ACI)
     ///
@@ -811,7 +811,7 @@ pub trait PushService: MaybeSend {
         &mut self,
         messages: OutgoingPushMessages,
     ) -> Result<SendMessageResponse, ServiceError> {
-        let path = format!("/v1/messages/{}", messages.recipient.uuid);
+        let path = format!("/v1/messages/{}", messages.destination);
         self.put_json(
             Endpoint::Service,
             &path,
diff --git a/libsignal-service/src/sender.rs b/libsignal-service/src/sender.rs
index 47c408722..c2218c6fa 100644
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -37,7 +37,7 @@ pub struct OutgoingPushMessage {
 
 #[derive(serde::Serialize, Debug)]
 pub struct OutgoingPushMessages {
-    pub recipient: ServiceAddress,
+    pub destination: uuid::Uuid,
     pub timestamp: u64,
     pub messages: Vec<OutgoingPushMessage>,
     pub online: bool,
@@ -498,7 +498,7 @@ where
                 .await?;
 
             let messages = OutgoingPushMessages {
-                recipient,
+                destination: recipient.uuid,
                 timestamp,
                 messages,
                 online,
diff --git a/libsignal-service/src/service_address.rs b/libsignal-service/src/service_address.rs
index caa22364d..4da35634e 100644
--- a/libsignal-service/src/service_address.rs
+++ b/libsignal-service/src/service_address.rs
@@ -1,7 +1,6 @@
 use std::convert::TryFrom;
 
 use libsignal_protocol::{DeviceId, ProtocolAddress};
-use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
 pub use crate::push_service::ServiceIdType;
@@ -15,7 +14,7 @@ pub enum ParseServiceAddressError {
     NoUuid,
 }
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize, Hash)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub struct ServiceAddress {
     pub uuid: Uuid,
     pub identity: ServiceIdType,
diff --git a/libsignal-service/src/websocket/sender.rs b/libsignal-service/src/websocket/sender.rs
index 166d9c800..e9ea90ef6 100644
--- a/libsignal-service/src/websocket/sender.rs
+++ b/libsignal-service/src/websocket/sender.rs
@@ -12,7 +12,7 @@ impl SignalWebSocket {
         &mut self,
         messages: OutgoingPushMessages,
     ) -> Result<SendMessageResponse, ServiceError> {
-        let path = format!("/v1/messages/{}", messages.recipient.uuid);
+        let path = format!("/v1/messages/{}", messages.destination);
         self.put_json(&path, messages).await
     }
 
@@ -21,7 +21,7 @@ impl SignalWebSocket {
         messages: OutgoingPushMessages,
         access: &UnidentifiedAccess,
     ) -> Result<SendMessageResponse, ServiceError> {
-        let path = format!("/v1/messages/{}", messages.recipient.uuid);
+        let path = format!("/v1/messages/{}", messages.destination);
         let header = format!(
             "Unidentified-Access-Key:{}",
             BASE64_RELAXED.encode(&access.key)

```

... although we could actually get rid of the `destination` field altogether (but so should Android!).

libsignal-service/src/profile_service.rs
```diff
@@ -22,7 +22,7 @@ impl ProfileService {
         let endpoint = match profile_key {
             Some(key) => {
                 let version = bincode::serialize(
-                    &key.get_profile_key_version(address.aci()),
+                    &key.get_profile_key_version(address.aci().unwrap()),
```
rubdos, line 25:
```suggestion
                    &key.get_profile_key_version(address.aci().expect("valid profile key version")),
```

I'd try to replace most of the unwraps in this PR with `expect`s

libsignal-service/src/push_service.rs
rubdos: Maybe the `uuid::Uuid` field should also be a `String`, such that we can have the `PNI:` prefix in there? I'm not sure how that works though, so that's your call. I don't think it'll matter in this case, because I expect the behaviour to be the same!

libsignal-service/src/service_address.rs
```diff
@@ -13,64 +15,124 @@ pub enum ParseServiceAddressError {
...
         &self,
         device_id: impl Into<DeviceId>,
     ) -> ProtocolAddress {
-        ProtocolAddress::new(self.uuid.to_string(), device_id.into())
+        match self.identity {
+            ServiceIdType::AccountIdentity => {
+                ProtocolAddress::new(self.uuid.to_string(), device_id.into())
+            },
+            ServiceIdType::PhoneNumberIdentity => ProtocolAddress::new(
+                format!("PNI:{}", self.uuid),
```
rubdos, line 34: We chop this prefix off in Whisperfish in the protocol store, right?

rubdos (commented):
This looks in a pretty good shape. Some minor things, and then let's get this in.

libsignal-service/src/push_service.rs
direc85: Your suggestion seems to work and it doesn't change anything at this point. Let's change it to `String` if and when the time comes.

libsignal-service/src/service_address.rs
direc85: It looks like we don't - but that's a discussion to be had in the Whisperfish MR.

PR #300, opened 2024-06-03
Add AccountManager::linked_devices

Schmiddiii:
This adds the functionality of decrypting `DeviceInfo::name` with the aci private key.

I'm unsure how much about this should live in libsignal-service-rs and how much in presage.
Feel free to suggest moving things around if needed.

See also https://github.com/whisperfish/presage/pull/193.

### DIFF
--- a/libsignal-service/src/account_manager.rs
+++ b/libsignal-service/src/account_manager.rs
@@ -8,9 +8,9 @@ use aes::cipher::{KeyIvInit, StreamCipher as _};
 use hmac::digest::Output;
 use hmac::{Hmac, Mac};
 use libsignal_protocol::{
-    kem, GenericSignedPreKey, IdentityKey, IdentityKeyStore, KeyPair,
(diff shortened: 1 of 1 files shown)


rubdos:
I would say this should live in the account-manager, next to the encrypt method, and it should probably take an `IdentityKeyPair` to be slightly more type safe. Moving it into account-manager allows you to implement some round-trip tests too!

Moreover, then we can add a method in the account manager to retrieve devices, and execute the decryption in there. We probably want to provide the AciIdentityStore to such a method.

Schmiddiii:
I have moved the functionality to AccountManager as wanted and added `AccountManager::linked_devices`.

I did not do round-trip tests, as all they would check is serialization/deserialization of DeviceName works (which is not something to really check) and `encrypt_device_name`/`decrypt_device_name` to be inverse, which is already checked by tests.

rubdos:
LGTM

PR #301, opened 2024-06-18
upgrade libsignal-client to v0.51.0

boxdot:


rubdos:
Looks good, @boxdot. Looks like you tested it too, so I'll pull it in.

PR #302, opened 2024-06-26
Add Configuration sync message sending

direc85:
Support sending messages with `SyncMessage { configuration: Option<Configuration> }` payload.

Tested on Whisperfish -- changing typing indicators setting synchronizes to Molly (which in turn is then able to send or not send typing indicator messages too).

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -657,6 +657,25 @@ where
         Ok(())
     }
 
+    /// Send `Configuration` synchronization message
+    #[tracing::instrument(skip(self))]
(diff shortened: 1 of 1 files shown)


rubdos:
I see we have the same for contact sync, but it would maybe be nice if we don't have to specify the recipient itself when calling this method. Anyway, good for now.

direc85:
> I see we have the same for contact sync, but it would maybe be nice if we don't have to specify the recipient itself when calling this method.

I was thinking about cleaning up that function call too -- I'll can do that along with the next PR.

PR #305, opened 2024-07-01
Install protobuf on pages CI

rubdos:


rubdos:
Guess this only is testable on `main`... #yolo

PR #309, opened 2024-07-14
Add `Keys` sync message

direc85:
Untested, but should work.

I tried to remove passing self-recipient, but couldn't figure out a way to get it in the method. I think that would require setting self service address to sender, which would break API, so I'm not delving into that currently.


direc85:
I'll rebase/recreate this on some older commit, since that makes the Whisperfish branch independent of the PNI branch (and all that jazz).

PR #310, opened 2024-07-14
Add `Keys` sync message

direc85:
Take two, intentionally (for development convenience) branched of an old commit - the one Whisperfish `main` currently uses.

I tried to remove passing self-recipient, but couldn't figure out a way to get it in the method. I think that would require setting self service address to sender, which would break API, so I'm not delving into that currently.

Real-world tests performed:

- [x] Linking Signal Desktop (tested with Whisperfish -> SD)
- [x] Responding to Signal Desktop `Keys` sync request
- [x] Sending a `Keys` sync request

### DIFF
--- a/libsignal-service/src/master_key.rs
+++ b/libsignal-service/src/master_key.rs
@@ -1,0 +1,111 @@
+const MASTER_KEY_LEN: usize = 32;
+const STORAGE_KEY_LEN: usize = 32;
+
+#[derive(Debug, PartialEq)]
+pub struct MasterKey {
+    pub inner: [u8; MASTER_KEY_LEN],
+}
+
+impl MasterKey {
+    pub fn generate() -> Self {
+        use rand::Rng;
+
+        // Create random bytes
+        let mut rng = rand::thread_rng();
+        let mut inner = [0_u8; MASTER_KEY_LEN];
+        rng.fill(&mut inner);
+        Self { inner }
+    }
+
+    pub fn from_slice(
+        slice: &Vec<u8>,
+    ) -> Result<Self, std::array::TryFromSliceError> {
+        let inner = slice.as_slice().try_into()?;
+        Ok(Self { inner })
+    }
+}
+
+impl From<MasterKey> for Vec<u8> {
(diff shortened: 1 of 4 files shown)


direc85:
Added WIP for MasterKey and StorageServiceKey, as well as MasterKeyStorage, which is probably done in some funky manner. Improvements and corrections welcome...

gferon:
Thanks! At the risk of missing the point, why was this suddenly needed? :smile:

direc85:
These are needed to properly support the Keys sync message, which is needed for linking nowadays... Ruben suggested making these into a trait, I'm not at all sure I did that in a good way.

Basically this part should be quite ready, but there's still the "where are these actually used" side of things...

direc85:
We still must test this code, but review-wise this PR can be considered ready.

libsignal-service/src/master_key.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        // Create random bytes
+        let mut rng = rand::thread_rng();
+        let mut inner = [0_u8; MASTER_KEY_LEN];
+        rng.fill(&mut inner);
+        Self { inner }
+    }
+
+    pub fn from_slice(
+        slice: &Vec<u8>,
```
rubdos, line 21:
```suggestion
        slice: &[u8],
```

libsignal-service/src/master_key.rs
```diff
@@ -0,0 +1,111 @@
...
+        let mut mac = HmacSha256::new_from_slice(&master_key.inner).unwrap();
+        mac.update(KEY);
+        let result = mac.finalize();
+        let inner: [u8; STORAGE_KEY_LEN] = result.into_bytes().into();
+
+        Self { inner }
+    }
+
+    pub fn from_slice(
+        slice: &Vec<u8>,
```
rubdos, line 56:
```suggestion
        slice: &[u8],
```

rubdos (changes requested):
Some generalizations, otherwise looking clean to me! :-)

rubdos (approved):

PR #312, opened 2024-07-22
feat: sync edited messages

boxdot:
This allows to send message edits and in particular syncs them when
using multi-device.

### DIFF
--- a/libsignal-service/src/sender.rs
+++ b/libsignal-service/src/sender.rs
@@ -331,8 +331,8 @@ where
         }
 
         // try to send the original message to all the recipient's devices
-        let mut results = vec![
-            self.try_send_message(
+        let result = self
+            .try_send_message(
                 *recipient,
                 unidentified_access.as_ref(),
                 &content_body,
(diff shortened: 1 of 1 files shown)


libsignal-service/src/sender.rs
```diff
@@ -331,51 +331,57 @@ where
...
+            .await?;
         }
 
         if end_session {
             let n = self.protocol_store.delete_all_sessions(recipient).await?;
             tracing::debug!("ended {} sessions with {}", n, recipient.uuid);
         }
 
-        results.remove(0)
+        result
```
boxdot, line 384: Maybe we should add the sync result here too, like it is done in send to group?

rubdos: I think that's the reason `results` used to be a Vec, but at this point I'm not sure anymore.

boxdot: This would break the result type of the public API. If that's ok, I can change it to returning a vec.

rubdos: I'd like for @direc85's opinion too, since he last touched these things iirc, but I think it makes sense. Maybe we should cross-check with what SA/SD do.

gferon: Every time we touch this part of the code we improve something and break another 😂 this is just bound to happen at this point. Hope your fix is the final one.

direc85: I'm in favor of returning a `Vec` here. What to do with the results is another question... Is it libsignals problem? Or the clients'?

direc85 (approved):
As long as we're allowed to send combo edit+data message, LGTM! I can't test this myself though until next week.

We're breaking the API with returning a `Vec` but it's better to have it this way.

libsignal-service/src/sender.rs
boxdot: I checked the impl in Signal Android. They only return the results of the actual message sending not the result of the sync message. In particular, when sending a message to a single recipient there is a single result, when sending to a group there is a list of results.

I think we should keep it aligned. I will remove the sync result from the list of result when sending to a group.

PR #314, opened 2024-08-02
Add destination ServiceAddress to Metadata

direc85:
What it says on the tin!

This is currently required only for delivering the destination ServiceAddress to the host application for determining our identity which the message was sent to in order to pick the correct protocol store when deleting all sessions.


rubdos (approved):
I think it makes sense to plumb it through, although we can also just extract it right before and pass it through the envelope handling ourselves at the application side.

@gferon/@boxdot, anything against this?

PR #315, opened 2024-08-12
fix: switch rustls backend back to ring

boxdot:
rultls switched their crypto backend to aws-lc, which unfortunately supports fewer platforms. For now, we switch the backend back to ring.

Also disable tls1.2 (signal servers supports tls1.3), and tokio native certs that are not used.

### DIFF
--- a/libsignal-service-hyper/Cargo.toml
+++ b/libsignal-service-hyper/Cargo.toml
@@ -22,7 +22,7 @@ url = "2.1"
 
 hyper = "1.0"
 hyper-util = { version = "0.1", features = ["client", "client-legacy"] }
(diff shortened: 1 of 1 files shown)


gferon (approved):

direc85:
LGTM -- this makes Whisperfish compile locally in Sailfish SDK for armv7hl again.

direc85:
@rubdos also LGTM'd this -- merging as discussed over at Signal.

direc85:
ntninja wrote a good comment about resolving the remaining issues in getting aws-lc working in sfdk: https://gitlab.com/whisperfish/whisperfish/-/merge_requests/593#note_2086577607 I'll drop the link here in case we need to make it work in the future.

PR #318, opened 2024-08-21
Add sending message request response messages

direc85:


direc85:
Sync-messages sent with this are accepted by Signal Desktop and mark the recipient blocked or unblocked. Only `Accept` and `Block` values were tested.

libsignal-service/src/sender.rs
```diff
@@ -691,6 +693,51 @@ where
         Ok(())
     }
 
+    /// Send `MessageRequestResponse` synchronization message with either a recipient ACI or a GroupV2 ID
+    #[tracing::instrument(skip(self))]
+    pub async fn send_message_request_response(
+        &mut self,
+        recipient: &ServiceAddress,
+        thread_aci: Option<&ServiceAddress>,
+        group_id: Option<&[u8; 64]>,
```
rubdos, line 702: Don't we have a stronger type for this?

libsignal-service/src/sender.rs
```diff
@@ -691,6 +693,51 @@ where
...
+        action: message_request_response::Type,
+    ) -> Result<(), MessageSenderError> {
+        match (&thread_aci, &group_id) {
```
rubdos, line 705: I'm okay with this for now, but sounds like we should have a `enum ThreadIdentifier { Aci(Uuid), Group(GroupID) }` for this

rubdos (approved):

libsignal-service/src/sender.rs
direc85: I looked for one but didn't find one. Not for the length either.

I think this is better resolved in another PR - I'll make an issue.

libsignal-service/src/sender.rs
direc85: I mirrored `MessageRequestResponse`  format but there's no need to. Especially so because the user can't click accept for a group and a recipient at the same time, so the check is better eliminated. I'll make it an enum.

PR #323, opened 2024-09-20
chore: fix clippy 1.81

boxdot:


libsignal-service/examples/storage.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::diverging_sub_expression)]
+
```
rubdos, line 2: Can you annotate the section where this is needed, instead of a global allow? If not, can you write a comment why this is here?

libsignal-service/src/messagepipe.rs
```diff
@@ -138,6 +138,7 @@ impl MessagePipe {
 /// WebSocketService that panics on every request, mainly for example code.
 pub struct PanicingWebSocketService;
 
+#[allow(clippy::diverging_sub_expression)]
```
rubdos, line 141: Here too, a comment explaining why it triggers, and why we suppress would be nice.

rubdos (changes requested):
Thanks :-)

rubdos:
@boxdot kind ping for the requested changes; maybe it wasn't clear from the notification that it was a change request :'-)

boxdot:
> @boxdot kind ping for the requested changes; maybe it wasn't clear from the notification that it was a change request :'-)

Thank you for reminding me. I got carried away.

rubdos:
Lint reasons don't seem to be stabilized yet in 1.75, any chance we can put the reason behind a nightly flag? Sounds a bit wrong to bump MSRV for this :'-)

(Sailfish is on 1.75 still, and will be for at least half a year I fear)

rubdos (approved):

PR #324, opened 2024-09-24
Fix deserialization of previous contacts

Schmiddiii:
This caused presage to not load contacts anymore stored from previous versions.

### DIFF
--- a/libsignal-service/src/models.rs
+++ b/libsignal-service/src/models.rs
@@ -16,6 +16,10 @@ pub struct Attachment<R> {
     pub reader: R,
 }
 
+const fn default_expire_timer_version() -> u32 {
+    1
+}
+
 /// Mirror of the protobuf ContactDetails message
 /// but with stronger types (e.g. `ServiceAddress` instead of optional uuid and string phone numbers)
 /// and some helper functions
@@ -29,6 +33,7 @@ pub struct Contact {
     pub verified: Verified,
     pub profile_key: Vec<u8>,
     pub expire_timer: u32,
+    #[serde(default = "default_expire_timer_version")]
     pub expire_timer_version: u32,
     pub inbox_position: u32,
     pub archived: bool,
--- a/libsignal-service/src/models.rs
+++ b/libsignal-service/src/models.rs
@@ -17,7 +17,7 @@ pub struct Attachment<R> {
 }
 
 const fn default_expire_timer_version() -> u32 {
-    1
+    2
 }
 
 /// Mirror of the protobuf ContactDetails message


rubdos:
I'm not 100% convinced that this is right: possibly we need to have this as `Option`al, and actually handle the absence of the timer version? There's quite some migration logic to this field.

Schmiddiii:
I don't know how exactly timer versions work, but I would expect the only way to handle a missing timer version is to start versioning the timer with version `1`.

rubdos:
I'm mostly wondering about what should happen if the timer is unset: should some messages be transmitted to actually trigger the creation of a timer version?

I haven't looked at this yet, but I've added it to my TODOs for my MR for WF: https://gitlab.com/whisperfish/whisperfish/-/merge_requests/627

Schmiddiii:
Looking at the migration [Signal Desktop](https://github.com/signalapp/Signal-Desktop/blob/main/ts/sql/migrations/1150-expire-timer-version.ts) has, they don't seem to ssend any messages.
Just set any future contacts timeouts to `1` and the current timeouts to `2` (not sure why `2`, maybe this MR value should also be changed to `2`).

direc85:
According to [this](https://github.com/signalapp/Signal-Android/commit/1f196f74ffeb15064d8d406b914ca183997159f4#diff-fb6b06bc4edddbfb57bdbed47b49be4e872618899299b0c9e721984a59dcc450R26), we should also check the recipient capabilities regarding the version. There are functions to update the timer with and without incrementing the timer.

So: 1 = recipient doesn't support the versioning, 2+ = recipient supports... Right?

Schmiddiii:
Given that the Signal servers now enforce that capability, its probably save to say that the recipient supports this.
And I even doubt that the choice of 1 vs 2 matters too much, if the recipient does not support that field, it will probably just be ignored when deserializing the protobuf.

Also, this MR is not for doing this absolutely correctly, this is just for fixing a regression where previous stores of presage do not work anymore, allowing applications to update presage in order to finally be able to link again (after now almost two weeks).

direc85:
I found [this](https://github.com/signalapp/Signal-Android/commit/1f196f74ffeb15064d8d406b914ca183997159f4#diff-01ca59555f4aae3c817ab6a0d3a92d99a5ceb97605962676aab94ffcd1be9abeR1513) which explains the groups situation - the version is always one.

rubdos:
> They also do not handle groups at all. I think expiration timers are not going to be versioned for groups. or at least not for the time being.

Groups are handled by the group state machine itself, which is already a versioning system on its own.

rubdos:
> If not, we need to keep every old version of all structs we use in the presage store forever for migration to work, or finally use a sensible storage backend.

You can probably handle most of this kind of migration logic internally. But yes, some more sensible storage method would be desirable. For now, I'd just say you put the struct you propose here in Presage, and make some From and Into implementations.

Schmiddiii:
Thinking about it some more, I don't think I actually need that patch anymore. I already shipped it out to my users once, their store should now have the new field. But I'll do something like that in the next breaking change to the store.

PR #329, opened 2024-10-12
Allow PNI in PendingMember

### LINKED ISSUE #328: Group decryption fails due PNI pending member
Background: there's a single group that did not show up in gurk until I filtered out errors in https://github.com/whisperfish/libsignal-service-rs/blob/65b51dbe7b42dae992f6a6d4c444bc2081a70be0/libsignal-service/src/groups_v2/operations.rs#L239-L243 . In particular, the group has three pending members. The decryption of one of those fails at https://github.com/whisperfish/libsignal-service-rs/blob/65b51dbe7b42dae992f6a6d4c444bc2081a70be0/libsignal-service/src/groups_v2/operations.rs#L134 with `GroupDecodingError::NotAci`. The group contains around 20 accounts that show up as "Deleted account" in the official client, which might be relevant in this case.

Tested on f99ff8324378ea653dbd5aa0c5d55e4414b34e23.

rubdos:
Should fix #328 

The other member decryption functions are ACI-only (which makes sense). I double-checked this.

@gferon, @boxdot, I have removed `Serialize` and `Deserialize` bounds on some structs. These were in place because Presage stores those structs directly in its database, but I don't think that is correct behavior. `libsignal-service` exposes those structs without any forward or backward compatibility guarantees, as illustrated in here: the `uuid` field is dropped in favor of the service_id field. Migration should be handled by the storage backend. Same reasoning goes for #324.

### DIFF
--- a/src/groups_v2/model.rs
+++ b/src/groups_v2/model.rs
@@ -5,6 +5,8 @@ use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 use zkgroup::profiles::ProfileKey;
 
+use crate::ServiceAddress;
+
 use super::GroupDecodingError;
 
 #[derive(Copy, Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
@@ -30,15 +32,15 @@ impl PartialEq for Member {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct PendingMember {
-    pub uuid: Uuid,
+    pub address: ServiceAddress,
     pub role: Role,
     pub added_by_uuid: Uuid,
     pub timestamp: u64,
 }
 
-#[derive(Derivative, Clone, Deserialize, Serialize)]
+#[derive(Derivative, Clone)]
 #[derivative(Debug)]
 pub struct RequestingMember {
     pub uuid: Uuid,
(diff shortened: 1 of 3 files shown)


rubdos:
Adding some more convenience methods and refactoring to this, while I'm implementing the changes in WF.

libsignal-service/src/groups_v2/model.rs
```diff
@@ -30,15 +31,15 @@ impl PartialEq for Member {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct PendingMember {
-    pub uuid: Uuid,
+    pub service_id: ServiceId,
```
rubdos, line 36: @gferon, what do you think, should we...

```suggestion
    pub service_id: ServiceAddress,
```

instead? I think maybe it's worth dropping `ServiceAddress` altogether, and sub it with `ServiceId` from proto...

gferon: The only problem is that `ServiceId` might be a bit more annoying to use? (to display for example). We can definitely try.

gferon (approved):

libsignal-service/src/groups_v2/model.rs
rubdos: We can postpone that until a future refactoring. I've subbed with serviceaddress now.

gferon:
> exposes those structs without any forward or backward compatibility guarantees

I just realized that while this statement is true, it's not really a big deal for data that's transient or can be re-fetched when invalidated (which is the case for groups).

rubdos:
> I just realized that while this statement is true, it's not really a big deal for data that's transient or can be re-fetched when invalidated (which is the case for groups).

Sure, but a big point of Signal is to rely as little as possible on state stored on the server. I don't think a "low-level" library like lss should facilitate that kind of "unorthodox behavior".

PR #330, opened 2024-10-13
Switch to attachments v4

### LINKED ISSUE #320: Attachment uploads to cdn0 are now rejected under latest Signal for ios
I noticed that Signal.app 7.26.1 (321) for ios started silently dropping my messages with attachments with this error:
```
2024/09/06 13:32:32:374  ❤ [AttachmentManagerImpl.swift:305 transitTierInfo(from:)]: assertionError: Invalid cdn info
2024/09/06 13:32:32:374  ❤ [MessageReceiver.swift:1144 processFlaglessDataMessage(_:request:thread:tx:)]: Could not build attachments!
```
Looking at the app code, it looks like ios no longer tolerates attachments that have been uploaded to cdn0:

https://github.com/signalapp/Signal-iOS/blob/7a19cb5ac898b8d5544ef3fae514b43de05182d5/SignalServiceKit/Messages/Attachments/V2/AttachmentManager/AttachmentManagerImpl.swift#L305-L308

As libsignal-server-rs only supports cdn0 currently, we'll need to add support for cdn2 and cdn3 for attachments to work going forward.

gferon:
which means dropping CDN0 which was AWS CloudFront and use CDN2 or CDN3 depending on what the server tells us to do.

TODO:
- [x] Make sure we still have `post_to_cdn0` functional, even though I'm not sure whether it is used or not.
- [x] Add logic for CDN2 (we cannot know whether it is still in use or not, the server tells us to use it).

In general, I'd like to ditch `libsignal-service-actix` since it is not used anymore, which in turn will let us remove the `PushService` trait and let us have a _much nicer_ direct access to the HTTP API.

Closes #320

### DIFF
--- a/src/push_service/cdn.rs
+++ b/src/push_service/cdn.rs
@@ -5,9 +5,9 @@ use std::{
 
 use futures::TryStreamExt;
 use reqwest::{
-    header::{CONTENT_LENGTH, CONTENT_TYPE},
+    header::{CONTENT_LENGTH, CONTENT_RANGE, CONTENT_TYPE, RANGE},
     multipart::Part,
-    Method,
+    Method, StatusCode,
 };
 use serde::Deserialize;
 use tracing::{debug, trace};
@@ -47,18 +47,6 @@ pub struct AttachmentDigest {
     pub digest: Vec<u8>,
     pub incremental_digest: Option<Vec<u8>>,
     pub incremental_mac_chunk_size: u64,
-}
-
-#[derive(Debug, Deserialize)]
-#[serde(rename_all = "camelCase")]
-pub struct ResumableUploadSpec {
-    attachment_key: Vec<u8>,
-    attachment_iv: Vec<u8>,
-    cdn_key: String,
-    cdn_number: u32,
-    resume_location: String,
-    expiration_timestamp: u64,
-    headers: HashMap<String, String>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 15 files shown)


libsignal-service-actix/src/push_service.rs
```diff
@@ -579,13 +579,14 @@ impl PushService for AwcPushService {
 
     async fn post_to_cdn0<'s, C: std::io::Read + Send + 's>(
         &mut self,
+        cdn_id: u32,
```
rubdos, line 582:
```suggestion
    async fn post_to_cdn<'s, C: std::io::Read + Send + 's>(
        &mut self,
        cdn_id: u32,
```

rubdos: (We probably want to retain upload capability to cdn0 for profile pics etc?)

libsignal-service-hyper/src/push_service.rs
```diff
@@ -413,6 +442,34 @@ impl PushService for HyperPushService {
...
     }
 
+    #[tracing::instrument(skip(self))]
+    async fn post(
+        &mut self,
+        service: Endpoint,
+        path: &str,
+        additional_headers: &[(&str, &str)],
+        credentials_override: HttpAuthOverride,
+    ) -> Result<HashMap<String, String>, ServiceError> {
```
rubdos, line 452: Post without any data?

gferon: yes : 😁

rubdos: In that case maybe
```suggestion
    #[tracing::instrument(skip(self))]
    async fn post_empty_body(
        &mut self,
        service: Endpoint,
        path: &str,
        additional_headers: &[(&str, &str)],
        credentials_override: HttpAuthOverride,
    ) -> Result<HashMap<String, String>, ServiceError> {
```
just for clarity?

rubdos: In that case maybe
```suggestion
    #[tracing::instrument(skip(self))]
    async fn post_empty_body(
        &mut self,
        service: Endpoint,
        path: &str,
        additional_headers: &[(&str, &str)],
        credentials_override: HttpAuthOverride,
    ) -> Result<HashMap<String, String>, ServiceError> {
```
just for clarity?

src/push_service/cdn.rs
```diff
@@ -90,26 +116,187 @@ impl PushService {
...
+    pub(crate) async fn get_attachment_resumable_upload_url(
         &mut self,
-        attrs: AttachmentV2UploadAttributes,
-        mut reader: impl Read + Send,
-    ) -> Result<(u64, Vec<u8>), ServiceError> {
-        let attachment_id = attrs.attachment_id;
-        let mut digester =
-            crate::digeststream::DigestingReader::new(&mut reader);
+        attachment_upload_form: &AttachmentUploadForm,
+    ) -> Result<Url, ServiceError> {
```
rubdos, line 123: The attachment_upload_form will be quite noisy in tracing; should we have some Display impl like I'm doing in #335 ?

src/push_service/cdn.rs
```diff
@@ -156,4 +342,98 @@ impl PushService {
...
+    async fn upload_to_cdn3(
+        &mut self,
+        resumable_url: &Url,
+        headers: &HashMap<String, String>,
+        content_type: &str,
+        content_length: u64,
+        mut content: impl std::io::Read + std::io::Seek + Send,
+    ) -> Result<AttachmentDigest, ServiceError> {
+        let resume_info = self
+            .get_attachment_resume_info_cdn3(resumable_url, &headers)
```
rubdos, line 392:
```suggestion
            .get_attachment_resume_info_cdn3(resumable_url, headers)
```

src/push_service/mod.rs
```diff
@@ -177,15 +177,14 @@ impl PushService {
         }
     }
 
-    #[tracing::instrument(skip(self, path), fields(path = %path.as_ref()))]
+    #[tracing::instrument(skip(self))]
     pub fn request(
         &self,
         method: Method,
         endpoint: Endpoint,
```
rubdos, line 184: Can we have some nice display implementation for `Endpoint`, for in tracing?

src/push_service/cdn.rs
```diff
@@ -156,4 +342,98 @@ impl PushService {
...
+    }
+
+    async fn upload_to_cdn3(
+        &mut self,
+        resumable_url: &Url,
+        headers: &HashMap<String, String>,
+        content_type: &str,
+        content_length: u64,
+        mut content: impl std::io::Read + std::io::Seek + Send,
+    ) -> Result<AttachmentDigest, ServiceError> {
```
rubdos, line 390: I would add tracing::instrument here, and ensure that the parameters are quite simple. We have some tracing events in the body.

rubdos (approved):
Some very minor suggestions (one is clippy, easily applied). This is very pretty code, well done.

src/push_service/cdn.rs
gferon: Just forgot 😄 will do.

src/push_service/cdn.rs
gferon: I think it's ok, it's not like we upload attachments like crazy anyways.

src/push_service/mod.rs
gferon: Done in 4e49b1bc29b1f8f8ce08bda6aecbcc384ec967a9

src/push_service/cdn.rs
rubdos: Maybe you don't! ;-)

rubdos (approved):

PR #331, opened 2024-10-15
Re-flatten crate structure and get rid of PushService trait

gferon:
Since the actix implementation is not used anymore, and makes changes to the HTTP layer harder, I'm suggesting this big refactoring. No logic has changed, and this is a pure copy-pasta from the contents of `libsignal-service-hyper`.

Among the benefits stand out:
- Better compilation error messages since we stop using `async_trait` (and static dispatch)
- Less generics type parameters propagated everywhere
- Our future ability to start using proper `http` types instead of having our own (weak) abstraction layer

This is a first step to refactor `PushService` and split the `push_service.rs` file in many smaller (and more readable) parts.

We can also start exposing different basic feature flags for toggling the TLS layer (maybe it is interesting to `whisperfish` to use OpenSSL?).

You can see the type of necessary changes downstream to `whisperfish` or `presage` (not that many) - https://github.com/whisperfish/presage/pull/272


.github/workflows/ci.yaml
```diff
@@ -37,18 +30,16 @@ jobs:
...
 
   build:
     name: Build (${{ matrix.project }}, Rust ${{ matrix.toolchain }})
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        project: ["libsignal-service-actix", "libsignal-service-hyper", "libsignal-service"]
+        project: ["libsignal-service"]
         toolchain: ["stable", "beta", "nightly"]
```
rubdos, line 41:
```suggestion
        toolchain: ["stable", "beta", "nightly", "1.75"]
```

README.md
```diff
@@ -46,20 +46,14 @@ We're actively trying to make `libsignal-service-rs` fully functional.
 
 If you're looking to contribute or want to ask a question, you're more than welcome to join our development channel on Matrix (#whisperfish:rubdos.be) or Libera.chat (#whisperfish) to get in touch with us!
 
-## Feature flags for libsignal-service
-
-| Feature flag     | Description                                                                                                                                                                             |
-| ---------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `unsend-futures` | This feature removes the `Send` requirement on returned futures. Enabling this flag may be necessary for interoperability with other libraries that don't support `Send` such as actix. |
-
```
rubdos, line 54: I'm not sure whether WF will be compatible for now without the Unsend feature.

gferon: Ah, I can definitely check that. It was my understanding that this was only for `libsignal-service-actix`. I'll double check.

gferon: with https://github.com/whisperfish/whisperfish/commit/e93ed195c0be420fe34a2b9b965412a58383fde9 I can confirm my initial suspicion, we can remove `unsend-futures`.

```
Checking whisperfish v0.6.0-dev (/Users/gferon/Developer/signal/whisperfish/whisperfish)
Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.52s
```

PR #332, opened 2024-10-16
Switch to reqwest

gferon:
Switch the concrete implementation of both the HTTP client to `reqwest` and `reqwest-websocket` which are powered behind the scenes by exactly the same libraries we are currently using (namely, `hyper` and `tokio-tungstenite`).

- Using a higher level builder-style API makes it easier to add support for new endpoints, and removes a lot of glue code we had to acquire over the year (the code looks now much closer to `PushService.kt`).
- `reqwest::Client` is now shared by both HTTP requests and the Websocket
- Remove one level of `sink` and `stream` in the websocket impl (since you don't have to forward messages to the concrete implementation)
- Introduce a no-op trait called `SignalServiceResponse` which is implemented for both `reqwest::Response` and `WebSocketResponseMessage` to let us share the HTTP status error handling (the one that handles mismatched sessions, etc.) - this is IMO a big improvement and will avoid any sort of drifting in the future


src/push_service/cdn.rs
```diff
@@ -32,170 +28,109 @@ pub struct AttachmentV2UploadAttributes {
...
+        form = form.text("x-amz-signature", upload_attributes.signature);
+
+        let mut buf = Vec::new();
+        reader
+            .read_to_end(&mut buf)
+            .expect("infallible Read instance");
+
+        form = form.text("Content-Type", "application/octet-stream");
+        form = form.text("Content-Length", buf.len().to_string());
+        form = form.part("file", Part::bytes(buf).file_name(filename));
```
direc85, line 123: I would keep the comment about the field ordering here.

direc85 (commented):
> +573 −1,050

_Everybody liked that_

(Except Clippy. Clippy has no chill.)

I gave this a quick read and it really seems to simmer the code down nicely!

gferon:
> > +573 −1,050
> 
> _Everybody liked that_
> 
> (Except Clippy. Clippy has no chill.)
> 
> I gave this a quick read and it really seems to simmer the code down nicely!

I have ruined my statistics a little bit by introducing a better abstraction for handling errors (which means we don't have duplicated logic in `PushService` and `WebSocketService` anymore 🎉) - however, I believe the tradeoff of a smaller statistical achievement for a higher quality library is worth it 🧌 

I think this is ready for review, I will not touch this PR again before it becomes an insatiable monster. I still believe it is quite OK to test and will be running `presage` on top of it for the next few days (the only part that I'm worried about would be attachments, but this can easily be tested.)

rubdos:
Haven't looked into detail too much yet, but this looks very good already. I suggest we also make a WF build with this and run it some days.

rubdos:
Whisperfish tests: https://gitlab.com/whisperfish/whisperfish/-/merge_requests/639

PR #334, opened 2024-10-18
Replace ServiceAddress by libsignal_protocol::ServiceId

gferon:
I hid a bunch of mistakes in there just to check if @rubdos is paying attention (jk, I think it should be fine).

This definitely changes _some_ things, like in some cases when preparing protos, you would get the prefixed `service_id_string` for PNI (which was the raw UUID before). I'm pretty sure it didn't use to work correctly.

Thank you for reading.


src/account_manager.rs
```diff
@@ -628,19 +628,19 @@ impl AccountManager {
     /// Should be called as the primary device to migrate from pre-PNI to PNI.
     ///
     /// This is the equivalent of Android's PnpInitializeDevicesJob or iOS' PniHelloWorldManager.
-    #[tracing::instrument(skip(self, aci_protocol_store, pni_protocol_store, sender, local_aci, csprng), fields(local_aci = %local_aci))]
+    #[tracing::instrument(skip(self, aci_protocol_store, pni_protocol_store, sender, local_aci, csprng), fields(local_aci =% local_aci.service_id_string()))]
```
rubdos, line 631:
```suggestion
    #[tracing::instrument(skip(self, aci_protocol_store, pni_protocol_store, sender, local_aci, csprng), fields(local_aci = %local_aci.service_id_string()))]
```

I think that's more standard notation?

src/profile_service.rs
```diff
@@ -1,46 +0,0 @@
...
-impl ProfileService {
-    pub fn from_socket(ws: SignalWebSocket) -> Self {
-        ProfileService { ws }
-    }
-
-    pub async fn retrieve_profile_by_id(
-        &mut self,
-        address: ServiceAddress,
-        profile_key: Option<zkgroup::profiles::ProfileKey>,
-    ) -> Result<SignalServiceProfile, ServiceError> {
```
rubdos, line 21: We're using this in WF; did it really just go away?

src/push_service/keys.rs
```diff
@@ -60,11 +61,14 @@ impl PushService {
...
+        destination: &ServiceId,
         device_id: u32,
     ) -> Result<PreKeyBundle, ServiceError> {
-        let path =
-            format!("/v2/keys/{}/{}?pq=true", destination.uuid, device_id);
+        let path = format!(
+            "/v2/keys/{}/{}?pq=true",
+            destination.raw_uuid(),
+            device_id
+        );
```
rubdos, line 71: IMO it would make sense if this had the domain separation; should we double check with Android? Matti may just know by heart.

src/push_service/keys.rs
```diff
@@ -83,13 +87,13 @@ impl PushService {
...
+        destination: &ServiceId,
         device_id: u32,
     ) -> Result<Vec<PreKeyBundle>, ServiceError> {
         let path = if device_id == 1 {
-            format!("/v2/keys/{}/*?pq=true", destination.uuid)
+            format!("/v2/keys/{}/*?pq=true", destination.raw_uuid())
         } else {
-            format!("/v2/keys/{}/{}?pq=true", destination.uuid, device_id)
+            format!("/v2/keys/{}/{}?pq=true", destination.raw_uuid(), device_id)
         };
```
rubdos, line 97: Owkay we have duplication of the pre key path now, maybe we want this deduped? Especially with the destination logic in there :-)

src/push_service/profile.rs
```diff
@@ -88,18 +89,17 @@ struct SignalServiceProfileWrite<'s> {
...
-            ))?;
+            let version =
+                bincode::serialize(&key.get_profile_key_version(address))?;
             let version = std::str::from_utf8(&version)
                 .expect("hex encoded profile key version");
-            format!("/v1/profile/{}/{}", address.uuid, version)
+            format!("/v1/profile/{}/{}", address.service_id_string(), version)
         } else {
-            format!("/v1/profile/{}", address.uuid)
+            format!("/v1/profile/{}", address.service_id_string())
```
rubdos, line 102: Here we *do* have the domain separation prefix; is this correct?

src/sender.rs
```diff
@@ -813,20 +818,20 @@ where
...
         })
     }
 
     // Equivalent with `getEncryptedMessages`
     #[tracing::instrument(
         level = "trace",
-        skip(self, unidentified_access, content, recipient),
-        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient),
+        skip(self, unidentified_access, content),
+        fields(unidentified_access = unidentified_access.is_some(), recipient = recipient.service_id_string()),
```
rubdos, line 830:
```suggestion
        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient.service_id_string()),
```

src/sender.rs
```diff
@@ -752,10 +757,10 @@ where
     }
 
     /// Send `Keys` synchronization message
-    #[tracing::instrument(skip(self))]
+    #[tracing::instrument(skip(self), fields(recipient = recipient.service_id_string()))]
```
rubdos, line 760:
```suggestion
    #[tracing::instrument(skip(self), fields(recipient = %recipient.service_id_string()))]
```

Using the display version is less verbose and more readable in my experience.

src/sender.rs
```diff
@@ -705,10 +710,10 @@ where
     }
 
     /// Send `MessageRequestResponse` synchronization message with either a recipient ACI or a GroupV2 ID
-    #[tracing::instrument(skip(self, recipient), fields(recipient = %recipient))]
+    #[tracing::instrument(skip(self), fields(recipient = recipient.service_id_string()))]
```
rubdos, line 713:
```suggestion
    #[tracing::instrument(skip(self), fields(recipient = %recipient.service_id_string()))]
```

src/sender.rs
```diff
@@ -686,10 +691,10 @@ where
     }
 
     /// Send `Configuration` synchronization message
-    #[tracing::instrument(skip(self, recipient), fields(recipient = %recipient))]
+    #[tracing::instrument(skip(self), fields(recipient = recipient.service_id_string()))]
```
rubdos, line 694:
```suggestion
    #[tracing::instrument(skip(self), fields(recipient = %recipient.service_id_string()))]
```

src/sender.rs
```diff
@@ -646,11 +651,11 @@ where
     /// Upload contact details to the CDN and send a sync message
     #[tracing::instrument(
         skip(self, unidentified_access, contacts, recipient),
-        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient),
+        fields(unidentified_access = unidentified_access.is_some(), recipient = recipient.service_id_string()),
```
rubdos, line 654:
```suggestion
        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient.service_id_string()),
```

src/sender.rs
```diff
@@ -496,11 +501,11 @@ where
     #[tracing::instrument(
         level = "trace",
         skip(self, unidentified_access, content_body, recipient),
-        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient),
+        fields(unidentified_access = unidentified_access.is_some(), recipient = recipient.service_id_string()),
```
rubdos, line 504:
```suggestion
        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient.service_id_string()),
```

src/sender.rs
```diff
@@ -931,12 +932,12 @@ where
     /// When no session with the recipient exists, we need to create one.
     #[tracing::instrument(
         level = "trace",
-        skip(self, unidentified_access, content, recipient),
-        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient),
+        skip(self, unidentified_access, content),
+        fields(unidentified_access = unidentified_access.is_some(), recipient = recipient.service_id_string()),
```
rubdos, line 936:
```suggestion
        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient.service_id_string()),
```

rubdos, line 936:
```suggestion
        skip(self, unidentified_access, content, recipient),
        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient.service_id_string()),
```

Shouldn't we skip the recipient in order to replace it?

src/sender.rs
```diff
@@ -1036,7 +1037,7 @@ where
                     } = sent;
                     UnidentifiedDeliveryStatus {
                         destination_service_id: Some(
-                            recipient.uuid.to_string(),
+                            recipient.service_id_string(),
```
rubdos, line 1040: Is this double checked with upstream?

rubdos (commented):
The strong types for Aci and Pni are awesome!  Clippy has a useful suggestion too ;-)

src/profile_service.rs
gferon: No, it's in `PushService`, and the impl was duplicated. Signal-Android really doesn't use the websocket for service API calls, so I think we should stick with it (especially since soon ™️  it will be using the same client behind the scenes).

src/sender.rs
gferon: The name of the field _really_ suggests that it should be like that.

src/push_service/keys.rs
gferon: By domain separation, you mean for ACI and PNI?

src/sender.rs
gferon: Looks like you don't have to, `tracing` doesn't complain about `recipient` not being `fmt::Display`.

src/push_service/keys.rs
gferon: How about dealing with those in another PR instead?

src/sender.rs
rubdos: > Note that overlap between the names of fields and (non-skipped) arguments will result in a compile error.

https://docs.rs/tracing/latest/tracing/attr.instrument.html#adding-fields

this is weeeeeird.

src/push_service/keys.rs
rubdos: Sure!

src/sender.rs
rubdos: Good enough!

src/sender.rs
rubdos: Well, it doesn't result in a compile error, so imo we can ship it 🤷

src/push_service/keys.rs
direc85: According to my [investigations](https://github.com/signalapp/libsignal/blob/826ee07b74666336f9eca74f2b270201ac157b37/rust/core/src/address.rs#L204), only ACI should have no prefix, and others (PNI) should have the corresponding (`pni:`) prefix.

src/push_service/profile.rs
direc85: IIUC, profiles are ACI-only (when someone contacts you by phone number, it's from-ACI-to-PNI message) so this looks correct to me.

What should be checked is how "I want to send a message to a phone number" is handled; IIRC there's no profile information fetched before at least sending the message...

src/sender.rs
```diff
@@ -526,7 +531,7 @@ where
                 .await?;
 
             let messages = OutgoingPushMessages {
-                destination: recipient.uuid,
+                destination: recipient.raw_uuid(),
```
direc85, line 534: Was this really raw-uuid-string only even before? Then again, I don't think we've ever supported ACI-to-PNI sending...

src/push_service/keys.rs
gferon: Yes, but does that apply to calls to these APIs endpoints?

src/push_service/profile.rs
gferon: Then to avoid the back-and-forth conversion, we can switch to `raw_uuid()`.

src/sender.rs
gferon: It might just work ™️ if we make the types fit correctly.

src/push_service/keys.rs
rubdos:
```java
      String path = String.format(PREKEY_DEVICE_PATH, destination.getIdentifier(), deviceSpecifier);
```
I'd say so.  I also don't find `pq=true` in the source code anymore.

src/sender.rs
```diff
@@ -343,12 +346,12 @@ where
 
     /// Send a message `content` to a single `recipient`.
     #[tracing::instrument(
-        skip(self, unidentified_access, message, recipient),
-        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient),
+        skip(self, unidentified_access, message),
+        fields(unidentified_access = unidentified_access.is_some(), recipient = recipient.service_id_string()),
```
rubdos, line 350:
```suggestion
        fields(unidentified_access = unidentified_access.is_some(), recipient = %recipient.service_id_string()),
```

src/sender.rs
rubdos: We still want display here, not the `impl Debug for String` one, I'd say.

src/service_address.rs
```diff
@@ -1,160 +1,39 @@
...
 
-    fn try_from(value: &str) -> Result<Self, Self::Error> {
-        if let Some(pni) = value.strip_prefix("PNI:") {
-            Ok(ServiceAddress::from_pni(Uuid::parse_str(pni)?))
-        } else {
-            Ok(ServiceAddress::from_aci(Uuid::parse_str(value)?))
+    fn aci(self) -> Option<Aci> {
+        match self.into() {
+            ServiceId::Aci(aci) => Some(aci),
+            ServiceId::Pni(_) => None,
```
rubdos, line 29: We can maybe consider upstreaming these two convenience methods.

rubdos: https://github.com/signalapp/libsignal/pull/595

src/sender.rs
gferon: This is never necessary when you deal with strings.

src/envelope.rs
```diff
@@ -85,30 +70,6 @@ impl Envelope {
         }
     }
 
-    fn new_from_entity(entity: EnvelopeEntity) -> Self {
```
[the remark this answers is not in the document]
gferon: This was actually unused after this refactoring, so I suppose.

src/account_manager.rs
```diff
@@ -641,19 +641,19 @@ impl AccountManager {
     /// Should be called as the primary device to migrate from pre-PNI to PNI.
     ///
     /// This is the equivalent of Android's PnpInitializeDevicesJob or iOS' PniHelloWorldManager.
-    #[tracing::instrument(skip(self, aci_protocol_store, pni_protocol_store, sender, local_aci, csprng), fields(local_aci = %local_aci))]
+    #[tracing::instrument(skip(self, aci_protocol_store, pni_protocol_store, sender, local_aci, csprng), fields(local_aci = %local_aci.service_id_string()))]
```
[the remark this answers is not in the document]
gferon: Sadly not, maybe we can make a PR in `libsignal`.

src/service_address.rs
gferon: Cool, thanks!

src/sender.rs
rubdos: Oh. Ok! :-)

gferon:
I think this is ready to 🚢 

Regarding your suggestions to use `recipent = %recipient.service_id_string()`, I'm not sure I exactly get it, because this is the impl of `tracing::Value` for `String`: https://docs.rs/tracing-core/0.1.32/src/tracing_core/field.rs.html#632 which means we don't need to call the `fmt::Display` wrapper one.

src/envelope.rs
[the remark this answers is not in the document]
gferon: Dropped! 💣

rubdos (approved):

PR #335, opened 2024-10-19
Log verbosity and sanitization

rubdos:
Implementing Display for several datatypes, for use in traces.

### DIFF
--- a/src/content.rs
+++ b/src/content.rs
@@ -124,7 +124,7 @@ impl fmt::Display for ContentBody {
                         )
                     },
                     (None, _, n) if n > 0 => {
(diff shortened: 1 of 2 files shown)


rubdos:
Would be cool to use valuable for some stuff, but I'm too lazy to research that decently atm.

gferon:
tbh it's a relatively shitty trait to implement. fmt::Display is good enough in most cases.

rubdos:
> tbh it's a relatively shitty trait to implement. fmt::Display is good enough in most cases.

Yeh indeed... Very much overkill. Merge this thing?

gferon (approved):

PR #336, opened 2024-10-19
Add missing pub use in groups_v2 module

gferon:


rubdos:
Hehe, does it show we don't store roles yet in WF? :face_with_peeking_eye:

gferon:
must be, because otherwise you're also unable to impl a conversion trait (or maybe with deep field access 😂).

PR #338, opened 2024-10-21
Only send a SyncMessage when sending to self

gferon:
I swore I'd never touch this part (because we broke it so many times) but here it is. I think this should be A-OK this time, though (famous last words).

Closes https://github.com/whisperfish/presage/issues/251


direc85 (approved):
Fixed in Whisperfish too!

PR #339, opened 2024-10-22
Fix sending note to self as primary without secondary devices

gferon:
There might be a better way to do this but... here's what happens:
1. Register a client as primary (no extra devices)
2. Try to send a message to `self` which will try to call `create_encrypted_message` but will never send any messages because we purposefully skip sending messages to ourselves (and own device) for good reasons
3. At the very end of the process, we fetch the `identity_key` (public key) we have for the recipient, which fails because we've never established a session with ourselves (this would only happen if we have other devices).

Here are two improvements:
- When the result of `create_encrypted_messages` return 0 messages, do nothing
- ...and check that we're `is_multi_device` before even attempting to craft and send a `SyncMessage` to self (that was missed when doing #338)

### DIFF
--- a/src/sender.rs
+++ b/src/sender.rs
@@ -131,6 +131,9 @@ pub enum MessageSenderError {
 
     #[error("Recipient not found: {addr:?}")]
     NotFound { addr: ServiceAddress },
+
+    #[error("no messages were encrypted: this should not really happen and most likely implies a logic error")]
+    NoMessagesToSend,
 }
 
 pub type GroupV2Id = [u8; GROUP_IDENTIFIER_LEN];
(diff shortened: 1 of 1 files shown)


src/sender.rs
```diff
@@ -553,10 +556,7 @@ where
                 // this can happen for example when a device is primary, without any secondaries
                 // and we send a message to ourselves (which is only a SyncMessage { sent: ... })
                 // addressed to self
-                warn!(
-                    "no messages were encrypted: this should not really happen and most likely implies a logic error."
-                );
-                break;
+                return Err(MessageSenderError::NoMessagesToSend);
```
rubdos, line 559: As far as I understand, with the new logic from #338, this code path can't be triggerend anymore by send-to-self?

gferon: indeed, hence why I wanna make it an error and we shall see if/when we break it.

rubdos: Sounds good! Merge ahead.

PR #342, opened 2024-11-03
[Partial revert] Make sure we use the provided CSPRNG everywhere.

gferon:
I just merged #341 but after talking with @boxdot the correct change is to do the opposite: make sure we use whatever the library user provide as CSPRNG.

There are only two places where we still use `rand::thread_rng`: in tests and when generating an ID for the next websocket request.


rubdos:
It makes sense both ways. By #331, we chose that we are an opinionated library. By being generic, we show we're flexible. Either way is fine by me.

rubdos:
@gferon conflicts here :-(

PR #345, opened 2024-11-16
Bump libsignal client to 0.62

rubdos:
This rewires some APIs to accept DeviceId instead of u32.

It also removes the ServiceIdExt trait, since
https://github.com/signalapp/libsignal/pull/595 got merged.  The `aci()` and `pni()` methods can be replaced with simple `into()` or `from()` calls, depending on context.


rubdos:
I made an attempt to make our `DEFAULT_DEVICE_ID` to be of `libsignal_protocol::DeviceId` type, instead of `u32`, but that isn't possible because the wrapped value is private, and `From::from` is not const. I could propose a `Default` implementation, or a custom `default` function upstream, to get rid of the weak type...

PR #346, opened 2024-11-18
Fix empty sync messages being sent

Schmiddiii:
This happened when giving `send_message` a `SyncMessage` to one self. In that case, the message would not be sent and instead `create_multi_device_sent_transcript_content` would create a new sync message, which is empty as the `SyncMessage` is neither a `DataMessage` nor an `EditMessage`.

This PR changes the condition to ignore sync messages.

This currently fails for me as I am rate limited.
Would be nice if someone not rate limited could try it out, otherwise I will wait for a while for the rate limit to stop.

### DIFF
--- a/src/sender.rs
+++ b/src/sender.rs
@@ -394,6 +394,7 @@ where
         }
 
         // don't send session enders as sealed sender
+        // sync messages are never sent as unidentified (reasons unclear), see: https://github.com/signalapp/Signal-Android/blob/main/libsignal-service/src/main/java/org/whispersystems/signalservice/api/SignalServiceMessageSender.java#L779
         if end_session || sync_message {
             unidentified_access.take();
         }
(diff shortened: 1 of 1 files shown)


Schmiddiii:
This does not seem to work yet.
Getting an error when syncing contacts:

```
SendFailed(
    MaximumRetriesLimitExceeded,
)
```

I  will need to look into it some more, but am already getting rate-limited after this failure happens once (even though the logs mention "sending via unidentified").

Schmiddiii:
The failure seems to be failing to create a session with device `2` for me.
I get the following log message three times when sending the sync message (after which, I get the rate limited error):

```
[2024-11-19T09:44:51Z DEBUG libsignal_service::sender] MismatchedDevices { missing_devices: [2], extra_devices: [] }
```

Not sure how to debug this further, but may try this afternoon.

Schmiddiii:
Figured out the issue, this message should not be sent with unidentified access.

I'm not sure if this should be with every SyncMessage (like implemented here) or if we should do that on the presage side.

PR #348, opened 2024-12-18
Catch HTTP 411 link error

rubdos:
Caught at https://gitlab.com/whisperfish/whisperfish/-/issues/724, untested.

### DIFF
--- a/src/push_service/response.rs
+++ b/src/push_service/response.rs
@@ -68,6 +68,27 @@ where
             })?;
             Err(ServiceError::ProofRequiredError(proof_required))
         },
+        StatusCode::LENGTH_REQUIRED => {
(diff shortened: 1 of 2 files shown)


src/push_service/response.rs
```diff
@@ -68,6 +68,27 @@ where
             })?;
             Err(ServiceError::ProofRequiredError(proof_required))
         },
+        StatusCode::LENGTH_REQUIRED => {
+            #[derive(Debug, serde::Deserialize)]
+            struct LinkedDeviceNumberError {
```
gferon, line 73: the problem of doing this is that if any other endpoint gets 411 we'll deserialize the response to this type, this has happened with `CONFLICT` before. I wonder what we could do that works better? maybe infer the possible custom error type with a template parameter?

PR #351, opened 2025-01-15
Use Auth header for ws login

rubdos:
Looks like upstream enforces this now, so all presage/lss/Whisperfish builds are currently broken. Testing as we speak.


rubdos:
This works. Merging because trivial, paging @gferon

PR #354, opened 2025-02-08
Fix profile demo code

direc85:
While rebasing usernames to main, `cargo test` told me that the example code for profile handling is wrong, so fix that. It got broken in 6401dc34872a5f0c2fba32c1f136d98ccf3dd418 I think.

And make Clippy happy once again.


rubdos:
(How don't we have doc comments in CI?)

PR #355, opened 2025-02-09
Add missing group change types

direc85:
Attempt to add the missing group v2 change message types. This is literally "it compiles so I'm gonna push it" so it's possible, likely even, that I'm doing a thing or two wrong. I just looked how the parsing works for the existing message types and extrapolated from there. This is somewhere between a draft PR and a RFC...

I was expecting the handling to be absent here, but the parsing was pretty much complete, just missing the newer additions. So this means that this is just the tip of the iceberg and the client has to handle all those messages. Independently of this PR, I can start working on Whisperfish already. Other clients are most welcome to join the improve-a-thon!

### DIFF
--- a/src/groups_v2/model.rs
+++ b/src/groups_v2/model.rs
@@ -1,7 +1,7 @@
 use std::{convert::TryFrom, convert::TryInto};
 
 use derivative::Derivative;
-use libsignal_protocol::ServiceId;
+use libsignal_protocol::{Aci, Pni, ServiceId};
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 use zkgroup::profiles::ProfileKey;
@@ -51,6 +51,34 @@ pub struct RequestingMember {
 impl PartialEq for RequestingMember {
     fn eq(&self, other: &Self) -> bool {
         self.uuid == other.uuid
+    }
+}
+
+#[derive(Derivative, Clone)]
+#[derivative(Debug)]
+pub struct BannedMember {
+    pub uuid: Uuid,
+    pub timestamp: u64,
+}
+
+impl PartialEq for BannedMember {
+    fn eq(&self, other: &Self) -> bool {
+        self.uuid == other.uuid
+    }
+}
+
+#[derive(Derivative, Clone)]
+#[derivative(Debug)]
+pub struct PromotedMember {
+    pub aci: Aci,
+    pub pni: Pni,
+    #[derivative(Debug = "ignore")]
+    pub profile_key: ProfileKey,
+}
+
+impl PartialEq for PromotedMember {
+    fn eq(&self, other: &Self) -> bool {
+        self.aci == other.aci && self.pni == other.pni
     }
 }
 
@@ -105,7 +133,6 @@ pub enum GroupChange {
         #[derivative(Debug = "ignore")]
         profile_key: ProfileKey,
     },
-    // for open groups
     NewPendingMember(PendingMember),
     DeletePendingMember(Uuid),
     PromotePendingMember {
@@ -113,23 +140,24 @@ pub enum GroupChange {
         #[derivative(Debug = "ignore")]
         profile_key: ProfileKey,
     },
-    // when admin control is enabled
+    Title(String),
+    Avatar(String),
+    Timer(Option<Timer>),
+    AttributeAccess(AccessRequired),
+    MemberAccess(AccessRequired),
+    InviteLinkAccess(AccessRequired),
     NewRequestingMember(RequestingMember),
     DeleteRequestingMember(Uuid),
     PromoteRequestingMember {
         uuid: Uuid,
         role: Role,
     },
-    // group metadata
-    Title(String),
-    Avatar(String),
-    Timer(Option<Timer>),
-    Description(Option<String>),
-    AttributeAccess(AccessRequired),
-    MemberAccess(AccessRequired),
-    InviteLinkAccess(AccessRequired),
     InviteLinkPassword(String),
+    Description(Option<String>),
     AnnouncementOnly(bool),
+    AddBannedMember(BannedMember),
+    DeleteBannedMember(Uuid),
+    PromotePendingPniAciMemberProfileKey(PromotedMember),
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
--- a/src/groups_v2/operations.rs
+++ b/src/groups_v2/operations.rs
@@ -2,7 +2,7 @@ use std::convert::TryInto;
 
 use base64::prelude::*;
 use bytes::Bytes;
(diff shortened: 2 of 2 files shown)


src/groups_v2/model.rs
```diff
@@ -54,6 +54,34 @@ impl PartialEq for RequestingMember {
     }
 }
 
+#[derive(Derivative, Clone)]
+#[derivative(Debug)]
+pub struct BannedMember {
+    pub uuid: Uuid,
```
rubdos, line 60: Should this be Aci?

src/groups_v2/model.rs
```diff
@@ -105,31 +133,31 @@ pub enum GroupChange {
         #[derivative(Debug = "ignore")]
         profile_key: ProfileKey,
     },
-    // for open groups
     NewPendingMember(PendingMember),
     DeletePendingMember(Uuid),
     PromotePendingMember {
         uuid: Uuid,
```
rubdos, line 139: Are pending members `Uuid`s, `ServiceId`s or `Aci`s? Maybe it's worth to comment the code whenever `Uuid` or `ServiceId` is used, to document *why* that specific form is used.

E.g., for invites, IIRC it's a `ServiceId`, because members can be invited based on their phone number (iirc).

src/groups_v2/model.rs
```diff
@@ -105,31 +133,31 @@ pub enum GroupChange {
...
     Avatar(String),
     Timer(Option<Timer>),
-    Description(Option<String>),
     AttributeAccess(AccessRequired),
     MemberAccess(AccessRequired),
     InviteLinkAccess(AccessRequired),
+    NewRequestingMember(RequestingMember),
+    DeleteRequestingMember(Uuid),
+    PromoteRequestingMember {
+        uuid: Uuid,
```
rubdos, line 152: Same question

rubdos:
@direc85 this is being tested, right? So maybe we should prioritize getting this merged? :-)

direc85:
> this is being tested, right? So maybe we should prioritize getting this merged? :-)

Not yet actually. Two reasons: this is pretty much a somewhat-incorrect WIP (types need to be verified) and it's a nuisance to carry a dirty `Cargo.toml` between branches. I'm focusing on the Whisperfish part currently, but if someone wants to contribute, feel free to pick this up, I really don't mind :)

src/groups_v2/model.rs
direc85: I think so yes. Parts of the code path accept ServiceId when creating BannedMember (MemberBanned) in SA, but in all concrete cases I could only find use of UUIDs or ACIs.

src/groups_v2/model.rs
direc85: Pending member is ServiceId indeed.

src/groups_v2/model.rs
direc85: This is Aci, by SA sources. This makes sense, because if I request a group membership, I'll start the process directly with Aci, w.r.t. if someone is being invited, it may or may not be a known recipient.

direc85:
Worked this further; still likely not complete. Also, even less tested.

rubdos:
@direc85 how close to mergable is this?

direc85:
> @direc85 this is being tested, right? So maybe we should prioritize getting this merged? :-)

Yes, I'm running this now nad have tested some group changes as well (add member, delete member).

> @direc85 how close to mergable is this?

Quite close; I'll still go through the types once (today-ish), and I'd like someone else have a look as well :)

src/groups_v2/operations.rs
```diff
@@ -84,6 +88,19 @@ impl GroupOperations {
...
+    fn decrypt_pni(
+        &self,
+        ciphertext: &[u8],
+    ) -> Result<Pni, GroupDecodingError> {
+        match self
+            .group_secret_params
+            .decrypt_service_id(bincode::deserialize(ciphertext)?)?
+        {
+            ServiceId::Pni(pni) => Ok(pni),
+            ServiceId::Aci(_aci) => Err(GroupDecodingError::NotPni),
```
rubdos, line 100: It could be useful to already throw a `tracing::error` in here, since this is behavior that shouldn't happen and probably indicates a bug somewhere.

I could imagine that a pni-only field gets a separate encoding (e.g. without any service-id indicator) some day, and that we don't catch this. The same probably goes for ACI.

rubdos (approved):
This looks rather clean!

src/groups_v2/operations.rs
direc85: Done!

direc85:
> TODO:
>     * `Member`: Uuid -> Aci, add Pni

Done, but adding Pni failed, because it's not in proto. Perhaps my assumption of lss `Member` == proto `DecryptedMember` is wrong, or we have to update protos or something? @rubdos any ideas?

>     * `PendingMember` Uuid -> Aci

Done.

>     * `BannedMember` Aci -> ServiceId

Done.

direc85:
After reading even more SA code, I ended up changing more existing types (mostly Uuid &rarr; Aci).  There was quite a lot of changes since the last ack, so a proper re-review is in order.

I also added sending Blocked message request response, but made the executive decision to ignore blocked GroupV1's. Not even sure if we have the blocked attribute; GroupV1 is sufficiently dead that this can be safely skipped. Whisperfish for example didn't support the feature at all.

rubdos (commented):
Ack on ignoring the GV1 things, those are not useful anymore anyway.

src/groups_v2/model.rs
```diff
@@ -27,30 +27,83 @@ pub struct Member {
...
+        s.serialize_str(&p.service_id_string())
+    }
+
+    pub fn deserialize<'de, D>(d: D) -> Result<Aci, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        // We have to go through String deserialization,
+        // because Aci does not implement Deserialize (duh).
+        let s = String::deserialize(d)?;
```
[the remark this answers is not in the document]
direc85: Perhaps, will check.

src/account_manager.rs
```diff
@@ -902,7 +902,6 @@ impl AccountManager {
     }
 }
 
-#[expect(clippy::result_large_err)]
```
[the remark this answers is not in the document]
direc85: Commit message says: `Rust 1.75 doesn't like these.`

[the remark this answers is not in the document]
direc85: I can drop this commit sure, but then we'll have to whisperfork lss again... Thoughts?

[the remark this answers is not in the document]
rubdos: I think we'll have to bump Rust either case for some recent changes :-(

Maybe that's some work for this summer.

PR #356, opened 2025-03-21
Bugfix: Include auth headers for attachment uploads.

ccstolley:
Yesterday attachment uploads started failing for me with error 401. I took a wild guess that the auth headers are now required in the upload request, and that appears to have worked.

I also added an `error_for_status()` check so that these types of errors aren't silently ignored in the future.

### DIFF
--- a/src/push_service/cdn.rs
+++ b/src/push_service/cdn.rs
@@ -426,9 +426,11 @@ impl PushService {
         .header("Upload-Offset", resume_info.content_start)
         .header("Upload-Length", buf.len())
         .header(CONTENT_TYPE, content_type)
+        .header("Authorization", headers["Authorization"].clone())
         .body(buf)
         .send()
-        .await?;
+        .await?
+        .error_for_status()?;
 
         trace!("attachment uploaded");
 
--- a/src/push_service/cdn.rs
+++ b/src/push_service/cdn.rs
@@ -417,20 +417,25 @@ impl PushService {
 
         trace!("digested content");
 
-        self.request(
+        let mut request = self.request(
             Method::PATCH,
             Endpoint::cdn(3, resumable_url.path()),
             HttpAuthOverride::Unidentified,
-        )?
-        .header("Tus-Resumable", "1.0.0")
-        .header("Upload-Offset", resume_info.content_start)
-        .header("Upload-Length", buf.len())
-        .header(CONTENT_TYPE, content_type)
-        .header("Authorization", headers["Authorization"].clone())
-        .body(buf)
-        .send()
-        .await?
-        .error_for_status()?;
+        )?;
+
+        for (key, value) in headers {
+            request = request.header(key, value);
+        }
+
+        request
+            .header("Tus-Resumable", "1.0.0")
+            .header("Upload-Offset", resume_info.content_start)
+            .header("Upload-Length", buf.len())
+            .header(CONTENT_TYPE, content_type)
+            .body(buf)
+            .send()
+            .await?
+            .error_for_status()?;
 
         trace!("attachment uploaded");
 
--- a/src/push_service/cdn.rs
+++ b/src/push_service/cdn.rs
@@ -275,7 +275,6 @@ impl PushService {
         &mut self,
         cdn_id: u32,
         resumable_url: &Url,
-        content_type: &str,
         content_length: u64,
         headers: HashMap<String, String>,
         content: impl std::io::Read + std::io::Seek + Send,
@@ -287,7 +286,6 @@ impl PushService {
             self.upload_to_cdn3(
                 resumable_url,
                 &headers,
-                content_type,
                 content_length,
                 content,
             )
@@ -386,7 +384,6 @@ impl PushService {
         &mut self,
         resumable_url: &Url,
         headers: &HashMap<String, String>,
-        content_type: &str,
         content_length: u64,
         mut content: impl std::io::Read + std::io::Seek + Send,
     ) -> Result<AttachmentDigest, ServiceError> {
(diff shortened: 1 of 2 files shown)


src/push_service/cdn.rs
```diff
@@ -426,9 +426,11 @@ impl PushService {
         .header("Upload-Offset", resume_info.content_start)
         .header("Upload-Length", buf.len())
         .header(CONTENT_TYPE, content_type)
+        .header("Authorization", headers["Authorization"].clone())
         .body(buf)
         .send()
-        .await?;
+        .await?
+        .error_for_status()?;
```
rubdos, line 433: As far as I recall, the `error_for_status` only really is meant to be used with the Signal API itself, so this might mis-map errors to internal types here. @gferon/@direc85 what's your opinion on that? I haven't worked on the new CDN code here, so Gabriel would be best qualified to make that call.

rubdos (commented):
I heard a whisper that @direc85 is also looking at this :-)

src/push_service/cdn.rs
direc85: I went through the code, and the only handling I can find is "ok-or-not" which leads into "attachment pointer or retry", so I think we should return "untranslated" errors here.

gferon: Agreed! IIRC you don't interact with Signal-server when uploading a blob to the CDN.

direc85: Sorry, the above is wrong. `error_for_status` makes sense here, but `service_error_for_status` would be wrong. This makes sense here!

rubdos: > Agreed! IIRC you don't interact with Signal-server when uploading a blob to the CDN.

In that case, the authorization header also sounds off. I can't imagine Signal using the same auth header for their own server as for CDN, that kinda violates some separation principles.

rubdos: @ccstolley: @direc85 also handled some stuff in #357, maybe look what happened there before double-investigating :-)

direc85: > I can't imagine Signal using the same auth header for their own server as for CDN, that kinda violates some separation principles.

Now that you mentioned that, I also noticed that my tests were all for CDN2 which was "always working". Also while being afk I remembered that the uploading process starts with the client asking *Signal servers* the details of where to upload and how; I bet that info has the headers necessary to work with the CDN servers. Combined, yeah, I'm changing my sides, and I don't think it makes sense to have the authentication here. Regarding Signal Android code, there's a code path that doesn't send the auth headers either, so that's at least plausible.

And, FWIW, there has been some issues with Signal services lately. That could mean that all this has been a coincidence.

@ccstolley I didn't really do much in the pull request, just double-thunk `[service_]error_for_status` usage.

In any case, thank you for the contribution and getting the ball rolling!

rubdos: > And, FWIW, there has been some issues with Signal services lately. That could mean that all this has been a coincidence.

Don't think that's true. They might have had issues on one CDN, and now everyone possibly points to CDN2 or something like that as a workaround.

ccstolley: I can confirm that cdn3 works with my patch, but will fail with 401 without it. Also just adding a properly encoded Authorization header with a bogus password will cause a 400.

This appears to be consistent with the [iOS version](https://github.com/signalapp/Signal-iOS/blob/main/SignalServiceKit/Upload/UploadEndpointCDN3.swift#L86-L89) which takes all of the response headers from the upload attribute request and passes them onto the actual cdn3 upload request (including presumably `Authorization`).

gferon: You're right! It seems to have been an omission when I first implemented it. We should then most likely do exactly that and use **all** of the headers provided as parameter?

ccstolley: @gferon I think that's probably the right fix, yes. Although personally I'd prefer something more explicit, one could imagine other headers suddenly becoming required in the future and we'd be back here debugging it again. I'll make that change.

gferon: Sorry, I was too quick 🙈 I pushed a commit in your branch already.

src/push_service/cdn.rs
```diff
@@ -417,18 +417,25 @@ impl PushService {
...
+
+        for (key, value) in headers {
+            request = request.header(key, value);
+        }
+
+        request
+            .header("Tus-Resumable", "1.0.0")
+            .header("Upload-Offset", resume_info.content_start)
+            .header("Upload-Length", buf.len())
+            .header(CONTENT_TYPE, content_type)
```
ccstolley, line 434: It's probably ok as-is but the other clients hard-code `content-type`, eg https://github.com/signalapp/Signal-iOS/blob/main/SignalServiceKit/Upload/UploadEndpointCDN3.swift#L87 . The [TUS](https://tus.io/protocols/resumable-upload#patch) protocol apparently mandates it as such for PATCH requests.

gferon: Very good point. I really wonder why we even have this parameters in the first place 🤔

gferon (approved):

gferon:
Just tested with `presage-cli`, uploading works again. Thanks a lot for figuring it out and sending a patch!

src/push_service/cdn.rs
rubdos: We can always just add the header list to a trace! call, although I'd see that as a security issue as well.

src/push_service/cdn.rs
```diff
@@ -417,18 +414,25 @@ impl PushService {
...
-        .header("Upload-Length", buf.len())
-        .header(CONTENT_TYPE, content_type)
-        .body(buf)
-        .send()
-        .await?;
+        )?;
+
+        for (key, value) in headers {
+            request = request.header(key, value);
+        }
```
rubdos, line 425: Possibly better to use [`headers`](https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.headers) here.

src/push_service/cdn.rs
direc85: @ccstolley Happy to be wrong! Passing the headers in full from upload form _sounds_ like the correct thing to do. Thank you for the contribution!

PR #357, opened 2025-03-22
Don't handle CDN errors as service errors

direc85:
While reviewing #356 I noticed that using `error_for_status` and `service_error_for_status` were used inconsistently. This PR tries to fix it. Rationale: when we're talking to a CDN, it doesn't make sense to translate e.g. HTTP 404 into `ServiceError::NotFound` which in some places can trigger (or could, rather) marking a recipient as unregistered. To me, this does not make any sense, so instead always return the HTTP error wrapped in `ServiceError::Http` to underline that we're _not_ talking with Signal servers with the request. We could also add an `ServiceError::CdnError` and add return the error there, but that's beyond of this PR.

Includes #356 as it is now. I'll drop the commit if/when that gets merged. Until then this is a draft.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -135,13 +135,12 @@ where
                 }
             }
 
-            if message.sync_message.is_some() {
-                if plaintext.metadata.sender.aci().map(Into::into)
+            if message.sync_message.is_some()
+                && plaintext.metadata.sender.aci().map(Into::into)
                     != Some(self.local_uuid)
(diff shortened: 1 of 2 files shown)


rubdos:
> Includes #356 as it is now. I'll drop the commit if/when that gets merged. Until then this is a draft.

Was merged.

rubdos (approved):
This makes a lot of sense.

PR #360, opened 2025-04-30
Pre-key consistency

rubdos:
The server view and client view on which prekeys are available *might* diverge over time, because of bugs and transmission failures. This PR tries to address this at least in part.

This PR also allows the client to assert that their last resort ("repeated use") keys are up-to-date and in sync, which is crucial if we don't want the device to get locked (https://github.com/signalapp/Signal-Server/commit/cdb651b68fc42447e765157c6311cd10f93828c6)

### DIFF
--- a/src/account_manager.rs
+++ b/src/account_manager.rs
@@ -183,6 +183,11 @@ impl AccountManager {
         {
             if protocol_store.signed_pre_keys_count().await? > 0
                 && protocol_store.kyber_pre_keys_count(true).await? > 0
+                && protocol_store.signed_prekey_id().await?.is_some()
+                && protocol_store
+                    .last_resort_kyber_prekey_id()
+                    .await?
+                    .is_some()
             {
                 tracing::debug!("Available keys sufficient");
                 return Ok(());
--- a/src/account_manager.rs
+++ b/src/account_manager.rs
@@ -172,6 +172,13 @@ impl AccountManager {
                 "Checking pre keys"
             ))
             .await?;
+        if !check_pre_keys {
+            tracing::info!(
+                "Last resort pre-keys are not up to date; refreshing."
+            );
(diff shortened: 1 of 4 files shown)


src/account_manager.rs
```diff
@@ -82,6 +82,51 @@ impl AccountManager {
...
+            protocol_store.last_resort_kyber_prekey_id().await?
+        else {
+            tracing::warn!("No last resort kyber prekey found");
+            return Ok(false);
+        };
+
+        let signed_prekey =
+            protocol_store.get_signed_pre_key(signed_prekey_id).await?;
+        let kyber_prekey =
+            protocol_store.get_kyber_pre_key(kyber_prekey_id).await?;
```
direc85, line 109: Out of scope, but it bugs me that the function names have `pre_key` instead of `prekey` :sweat_smile:

rubdos: We should indeed get those all in order. And we should change all the occurances of i32 in our own traits to the type safe versions provided by lsc.

src/push_service/keys.rs
```diff
@@ -45,6 +45,52 @@ impl PushService {
...
 
+    /// Checks for consistency of the repeated-use keys
+    ///
+    /// Supply the digest as follows:
+    /// `SHA256(identityKeyBytes || signedEcPreKeyId || signedEcPreKeyIdBytes || lastResortKeyId ||
+    /// lastResortKeyBytes)`
+    ///
+    /// The IDs are represented as 8-byte big endian ints.
+    ///
+    /// Retuns `Ok(())` if the view is consistent.
```
direc85, line 56: This is not quite correct; `Ok(true)` here I think, and what does `Ok(false)` mean?

gferon (approved):

PR #361, opened 2025-05-15
chore: upgrade to libsignal 0.72.1

boxdot:


src/account_manager.rs
```diff
@@ -294,11 +294,8 @@ impl AccountManager {
         let pub_key = query
             .get("pub_key")
             .ok_or(ProvisioningError::MissingPublicKey)?;
-        let pub_key = BASE64_RELAXED
-            .decode(&**pub_key)
-            .map_err(|e| ProvisioningError::InvalidPublicKey(e.into()))?;
-        let pub_key = PublicKey::deserialize(&pub_key)
-            .map_err(|e| ProvisioningError::InvalidPublicKey(e.into()))?;
+        let pub_key = BASE64_RELAXED.decode(&**pub_key)?;
```
gferon, line 297: I think we should still keep the specific error here?

src/provisioning/mod.rs
```diff
@@ -85,6 +86,10 @@ pub enum ProvisioningError {
     EncryptOnlyProvisioningCipher,
     #[error("invalid profile key bytes")]
     InvalidProfileKey(TryFromSliceError),
+    #[error(transparent)]
+    Curve(#[from] CurveError),
+    #[error(transparent)]
+    Base64(#[from] base64::DecodeError),
```
rubdos, line 92: Should we consolidate these three errors? I don't think we should expose low-level errors like Curve and base64 in the public API.

src/push_service/error.rs
```diff
@@ -98,4 +102,7 @@ pub enum ServiceError {
 
     #[error("HTTP reqwest error: {0}")]
     Http(#[from] reqwest::Error),
+
+    #[error(transparent)]
+    Curve(#[from] CurveError),
```
rubdos, line 107: Same here, sounds like we want a more "high level" kind of error.

src/account_manager.rs
```diff
@@ -751,17 +751,18 @@ impl AccountManager {
             HashMap::<String, u32>::with_capacity(local_device_ids.len());
 
         let signature_valid_on_each_signed_pre_key = true;
+        let default_device_id: u32 = (*DEFAULT_DEVICE_ID).into();
         for local_device_id in
-            std::iter::once(DEFAULT_DEVICE_ID).chain(local_device_ids)
+            std::iter::once(default_device_id).chain(local_device_ids)
```
rubdos, line 760: Should we instead make sure that `local_device_ids` is also the strong type, from the get_sub_device_sessions call? Or don't we control that trait?

src/envelope.rs
```diff
@@ -77,7 +77,7 @@ impl Envelope {
     }
 
     pub fn is_receipt(&self) -> bool {
-        self.r#type() == crate::proto::envelope::Type::Receipt
+        self.r#type() == crate::proto::envelope::Type::ServerDeliveryReceipt
```
rubdos, line 80: Should we rename our own method as well here? Think that'd make sense.

src/profile_cipher.rs
```diff
@@ -233,6 +235,32 @@ impl ProfileCipher {
     }
 }
 
+struct Rng06Shiv<'a, T>(&'a mut T);
```
rubdos, line 238: Either get rid of the 0.6, or document why we still need it around.

gferon:
> I think my previous comments still stand as well.

yeah, I've just fixed those :)

gferon:
@rubdos could you maybe help me figure out if https://github.com/whisperfish/libsignal-service-rs/pull/361/files#diff-b3f0f8b48dd131014e8b7edaee61b028099f4a17ff007f968ff4b532d22881beR644 is correct or not (I haven't checked for now, and I think it's the only risky part of this PR).

PR #362, opened 2025-05-24
refactor: don't store Rng in MessageSender

boxdot:
A `Rng` instance was stored in `MessageSender`, and was used to generate random numbers. However, this might cause a problem: Cloning `MessageSender` also cloned the RNG, leading to the generation of identical random numbers from multiple cloned instances. This is a security risk.

This commit removes the `Rng` from the `MessageSender` struct.  Random number generation is now handled on a per-function basis, using `thread_rng()` when needed. This ensures that each function call gets its own independent RNG state, preventing the previously observed issue.

Also see whisperfish/presage#292.

### DIFF
--- a/src/proto.rs
+++ b/src/proto.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::all)]
 
-use rand::{CryptoRng, Rng};
+use rand::{thread_rng, Rng, RngCore};
 include!(concat!(env!("OUT_DIR"), "/signalservice.rs"));
 include!(concat!(env!("OUT_DIR"), "/signal.rs"));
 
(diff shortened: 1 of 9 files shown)


src/proto.rs
```diff
@@ -69,10 +69,11 @@ impl WebSocketResponseMessage {
 }
 
 impl SyncMessage {
-    pub fn with_padding<R: Rng + CryptoRng>(csprng: &mut R) -> Self {
-        let random_size = csprng.gen_range(1..=512);
+    pub fn with_padding() -> Self {
+        let mut rng = thread_rng();
```
rubdos, line 73: Shouldn't we just provide the rng, like we do everywhere else in the API? I wouldn't depend on a specific Rng.

boxdot: Sure. Let me restore it

PR #366, opened 2025-06-16
Migrate PushService methods to the emulated HTTP request/response inside of the websocket

gferon:


rubdos:
@gferon yes please, this looks very good to me!

PR #371, opened 2025-09-06
Fix sync messages

direc85:
It turns out we have been creating and sending SyncMessages wrong; this PR fixes it:

- Don't try to send invalid SyncMessages during normal send
  - Only sync if there's a message or edit-message
- Add send_sync_message helper
  - takes SyncMessage, which now needs to be created on client side

The approach here is a bit blunt, but should crystalize what this is all about. I opted to remove all the send-sync-this-and-that helpers and replace it with a single helper which just sends the sync message *and does literally nothing else*.

This commit is untested, but [the one in whisperfork](https://github.com/whisperfish/libsignal-service-rs/commit/fec2606fe846e16ba86cf6c2844f8fae714fa2e6) has been tested as follows:

- changing use read receipts setting in WF syncs to SD
- reading a message in WF syncs to SD and marks the message/thread read as expected
- linking (relinking) SD to WF now happens quickly
- sending messages from WF syncs properly in SD

### DIFF
--- a/src/sender.rs
+++ b/src/sender.rs
@@ -377,23 +377,30 @@ where
         // only send a sync message when sending to self and skip the rest of the process
         if message_to_self && is_multi_device && !sync_message {
             debug!("sending note to self");
-            let sync_message = self
+            if let Some(sync_message) = self
                 .create_multi_device_sent_transcript_content(
                     Some(recipient),
                     content_body,
                     timestamp,
                     None,
-                );
-            return self
-                .try_send_message(
-                    *recipient,
-                    None,
-                    &sync_message,
-                    timestamp,
-                    include_pni_signature,
-                    online,
                 )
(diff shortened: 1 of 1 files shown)


src/sender.rs
```diff
@@ -377,23 +377,30 @@ where
         // only send a sync message when sending to self and skip the rest of the process
         if message_to_self && is_multi_device && !sync_message {
             debug!("sending note to self");
-            let sync_message = self
+            if let Some(sync_message) = self
```
rubdos, line 380: This logic makes a lot more sense to me :'-)

gferon:
great! only took us 5 years and many adjustments to get right 😅

rubdos (approved):
Removing lines and fixing bugs at the same time? Count me in.

PR #373, opened 2025-09-06
Fix renaming devices

direc85:
I had to change the function signature quite a bit to compensate for `async` shenanigans (the original signature required `IdentityKey` which was `async` and the function itself is `async`). If this is not acceptable, I can try to change it, but this seems to work fine.

Tested with Whisperfish; the changed name gets saved and is displayed again correctly.

### DIFF
--- a/src/account_manager.rs
+++ b/src/account_manager.rs
@@ -650,12 +650,23 @@ impl AccountManager {
     /// Update (encrypted) device name
     pub async fn update_device_name<R: Rng + CryptoRng>(
         &mut self,
+        device_id: u8,
         device_name: &str,
-        public_key: &IdentityKey,
+        aci: Aci,
+        aci_identity_store: &dyn IdentityKeyStore,
         csprng: &mut R,
     ) -> Result<(), ServiceError> {
+        let device_id = libsignal_core::DeviceId::new(device_id).unwrap();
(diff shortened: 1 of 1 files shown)


rubdos:
@direc85 some compilation failures here, probably to do with the mismatch in libsignal-core versions between fork and main?

direc85:
Yeah, that was it indeed! I didn't even pay attention to the "some checks failed" email because there have been failing checks every time I've made a pull request.

src/account_manager.rs
```diff
@@ -650,12 +650,23 @@ impl AccountManager {
...
+        aci_identity_store: &dyn IdentityKeyStore,
         csprng: &mut R,
     ) -> Result<(), ServiceError> {
+        let device_id = libsignal_core::DeviceId::new(device_id).unwrap();
+        let addr = aci.to_protocol_address(device_id).unwrap();
+        let public_key = aci_identity_store.get_identity(&addr).await?;
+        let Some(public_key) = public_key else {
+            return Err(ServiceError::SendError {
+                reason: format!("public key for device {addr:?} not found")
+                    .into(),
```
rubdos, line 665:
```suggestion
                reason: format!("public key for device {addr:?} not found"),
```

Can dump the `.into()` call, format!() is a string either way.

PR #377, opened 2025-10-08
Remove unused dependency

Schmiddiii:
As far as I can see, this dependency does nothing. There is one struct annotated with `serde_as`, but as far as I can see from the [serde_with documentation](https://docs.rs/crate/serde_with/latest), it would also require annotating fields to actually do something.

This was introduced in #361.


rubdos:
@boxdot since you introduced this, any comment?

PR #378, opened 2025-10-08
Fix deprecated SPDX license used

Schmiddiii:
According to https://spdx.org/licenses/, AGPL-3.0 is deprecated. Swapped this out by AGPL-3.0-only (there would also exists -or-later, but to me knowledge we are forced to use -only as that is also the license used by Signal: https://github.com/signalapp/libsignal/blob/eb7e0e7d18c76b29a0de706ed8932730f68b677e/Cargo.toml#L42).


rubdos:
Yep!

PR #394, opened 2026-02-20
Update protos

gferon:
Nothing to see here, just your regular update to track upstream.


rubdos (commented):
You'll want to try this again after the groupv2 stuff is merged (#401), because these bumps seem to influence the `operations.rs` a lot

PR #419, opened 2026-03-13
Revert to REST api for registration

thandal:
See https://github.com/whisperfish/presage/issues/371

(there is an accompanying change coming for presage-cli)

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -163,6 +163,37 @@ where
         }
     }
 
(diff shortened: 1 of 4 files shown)


gferon:
I'm pretty sure this isn't the solution because I have started to work on this, and you can get the code with the websocket and move forward. There were just a few mistakes made with the assumptions that it should work the same as before.

PR #426, opened 2026-04-06
unauth profile fetching and profile credentials

rubdos:
Needs additional changes to the AccountManager with fallback logic, and needs testing. Will do after releasing Whisperfish beta 40.


rubdos:
Cc @roder, this accidentally implements the Profile Credential fetching which you'll need for your GV2 work. I'll give your PR a review now...

Issue #209, opened 2023-03-25
PlaintextContent not handled


direc85:
Note: I practised some proper tomfoolery with my Signal Desktop database before this happened, so this may be a situation Signal isn't supposed to end up in. Reader discretion advised.

Using Whisperfish and Signal Desktop, I managed to have cross-device (sync?) messages sent as `PlaintextContent` type. Here's roughly how it went:

- Have Whisperfish open, linked Signal Desktop closed
- Send a message to self (or another contact)
- Open Signal Desktop
- &rarr; Lots of "A message from [my_name] couldn't be delivered" messages appear (one per sent message)
- None of the messages sent from WF to SD show up, the other direction works

Here's a snippet from Whisperfish log:

```
2023-03-25 17:46:52.917 [TRACE] libsignal_service::envelope: Envelope::decrypt: not encrypted
2023-03-25 17:46:52.917 [WARN] Unknown envelope type PlaintextContent
2023-03-25 17:46:52.918 [ERROR] Error opening envelope: InvalidFrameError { reason: "Envelope has unknown type PlaintextContent." }
```

After following the code for a bit, I found out that envelope type 8 is `PLAINTEXT_CONTENT` and that it is not matched in `libsignal-service/src/cipher.rs` function `decrypt(&mut self, envelope: Envelope)`, instead it falls to the default case:

```
_ => {
    // else
    return Err(ServiceError::InvalidFrameError {
        reason: format!(
            "Envelope has unknown type {:?}.",
            envelope.r#type()
        ),
    });
},
```

Signal logs look like this:

```
{"level":30,"time":"2023-03-25T18:05:44.194Z","msg":"MessageReceiver.decryptAndCacheBatch 1"}
{"level":30,"time":"2023-03-25T18:05:44.194Z","msg":"SignalProtocolStore.enterZone(decryptAndCacheBatch:MessageReceiver)"}
{"level":30,"time":"2023-03-25T18:05:44.195Z","msg":"queueing envelope [REDACTED]2ce.1 > [REDACTED]2ce 1679767534505 (REDACTED_SOMETHING)"}
{"level":30,"time":"2023-03-25T18:05:44.195Z","msg":"MessageReceiver.decryptEnvelope([REDACTED]2ce.1 > [REDACTED]2ce 1679767534505 (REDACTED_SOMETHING))"}
[...]

PR #423, opened 2026-03-20
No sender key state recovery: SealedSenderDecryptionError

rubdos:
This implements the discussion of https://github.com/signalapp/libsignal/pull/494

PR #428, opened 2026-04-22
fix: handle sealed-sender PlaintextContent in service cipher

### LINKED ISSUE #427: sealed_sender_decrypt rejects PlaintextContent → DecryptionErrorMessage envelopes are dropped, peers stuck in decryption loop
## Summary

`sealed_sender_decrypt_with_validated_usmc` in `src/cipher.rs` only handles three inner `CiphertextMessageType` variants — `Whisper`, `PreKey`, `SenderKey` — and falls through to `InvalidMessage(msg_type, "unexpected message type for sealed_sender_decrypt")` for `Plaintext`. Since `Plaintext` is the USMC wrapper Signal clients use to send a sealed-sender `DecryptionErrorMessage` (the protocol's session-restart mechanism), the envelope is dropped at the cipher layer and never reaches presage's `ContentBody::DecryptionErrorMessage` handler.

The consequence, observed in production against an otherwise-fine Signal Desktop / iOS peer: when crypto drift occurs (e.g. the peer's session state is out of sync after an app update), every subsequent outbound message fails on the peer's side, their client sends back a sealed-sender `DecryptionErrorMessage`, we log `error opening envelope ... Sealed sender decryption error: invalid Plaintext message: unexpected message type for sealed_sender_decrypt, usmc: true`, drop the reset request, and the cycle repeats indefinitely. The user sees "Viestiä ei voitu toimittaa" / "Message could not be delivered" on every outbound.

This is the sealed-sender analog of [#209](https://github.com/whisperfish/libsignal-service-rs/issues/209) — fixed for the non-sealed `Type::PlaintextContent` envelope path in [ce73b53](https://github.com/whisperfish/libsignal-service-rs/commit/ce73b53f6211046e749072f542e795b5f9273ef8) (`Implement PlaintextContent`).

## Location

https://github.com/whisperfish/libsignal-service-rs/blob/782c0d6/src/cipher.rs#L730-L767

```rust
let message = match usmc.msg_type()? {
    CiphertextMessageType::Whisper => { … }
    CiphertextMessageType::PreKey => { … }
    CiphertextMessageType::SenderKey => { … }
    msg_type => {
        return Err(SignalProtocolError::InvalidMessage(
            msg_type,
            "unexpected message type for sealed_sender_decrypt",
        ));
    },
};
```

## Fix direction

Add a `CiphertextMessageType::Plaintext` arm. The USMC contents for a Plaintext type are already in `PlaintextContent` wire format (`0xC0 | Content proto | 0x80 | zero-pad`). Strip the `0xC0` identifier byte and return the rest; the caller's existing `strip_padding` consumes the trailing `0x80` pad, leaving Content proto bytes that `Content::from_proto` already routes to `ContentBody::DecryptionErrorMessage`.

## Reproduction

Production log snippet from an `Sqlite`-backed presage deployment, same ACI pair over ~4 hours:

```
[peer sends DEM via sealed sender]
Sealed sender decryption error: invalid Plaintext message: unexpected message type for sealed_sender_decrypt, usmc: true
[we retry next outbound on stale ratchet]
Sealed sender decryption error: invalid Plaintext message: unexpected message type for sealed_sender_decrypt, usmc: true
Sealed sender decryption error: invalid Plaintext message: unexpected message type for sealed_sender_decrypt, usmc: true
…
```

## PR

#428 — adds the `Plaintext` arm, removes the now-unreachable catch-all, and adds a test that builds a real `PlaintextContent`-wrapped `DecryptionErrorMessage`, runs it through `sealed_sender_decrypt_with_validated_usmc`, and exercises the whole downstream pipeline (`strip_padding` → `Content::decode` → `Content::from_proto`) to assert the final body is `ContentBody::DecryptionErrorMessage`.

tow:
Fixes #427.

## Summary

`sealed_sender_decrypt_with_validated_usmc` previously rejected `CiphertextMessageType::Plaintext` with `"unexpected message type for sealed_sender_decrypt"`, dropping the sealed-sender `DecryptionErrorMessage` envelopes Signal clients use to ask for a session reset. Both parties then stayed stuck in a decryption-failure loop: our next outbound reused the stale ratchet, the peer sent another session-reset, we dropped that too.

This is the sealed-sender analog of #209 (fixed for the non-sealed `Type::PlaintextContent` path in ce73b53).

## Fix

Add a `Plaintext` arm that strips the `0xC0` `PlaintextContent` identifier byte. The caller's existing `strip_padding` removes the trailing `0x80` ISO7816 pad, leaving the Content proto bytes that the envelope pipeline already decodes via `Content::from_proto` — which already routes `decryption_error_message` to `ContentBody::DecryptionErrorMessage`.

With all four `CiphertextMessageType` variants now covered, the catch-all fallthrough becomes unreachable; drop it. The enum is not `#[non_exhaustive]` so removing the arm is sound.

## Test

The test builds a real `DecryptionErrorMessage`, wraps it in a `PlaintextContent`-framed USMC, decrypts via `sealed_sender_decrypt_with_validated_usmc`, then simulates every step `open_envelope()` runs afterwards — `strip_padding`, `Content::decode`, `Content::from_proto` — with `was_plaintext: false` as the real sealed-sender branch produces, and asserts the resulting body is `ContentBody::DecryptionErrorMessage`. That covers every step of the downstream pipeline that differs from the already-working non-sealed path.

## Test plan

- [x] `cargo test --lib sealed_sender_plaintext_usmc_returns_content_proto` — passes
- [x] Full lib test suite (16 tests) passes, no new warnings
- [x] Deployed against live Sanna-thread traffic; real-world behaviour documented in issue

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -758,11 +758,28 @@ async fn sealed_sender_decrypt_with_validated_usmc(
             group_decrypt(usmc.contents()?, sender_key_store, remote_address)
                 .await?
         },
-        msg_type => {
-            return Err(SignalProtocolError::InvalidMessage(
-                msg_type,
-                "unexpected message type for sealed_sender_decrypt",
-            ));
+        CiphertextMessageType::Plaintext => {
+            // Sealed sender envelope wrapping a PlaintextContent, used by
+            // recipients to send a DecryptionErrorMessage back to the sender
+            // when their local session state is out of sync and they need a
+            // new session established. Without this branch the envelope is
+            // dropped and the sender never learns to reset the session,
+            // leaving the two parties stuck in a decryption-failure loop.
+            //
+            // The PlaintextContent wire format is:
+            //   0xC0 (identifier) | Content proto encoding | 0x80 (pad) | 0x00*
+            // Strip the 0xC0 prefix here; the caller's strip_padding() removes
+            // the 0x80-terminated ISO7816 padding, leaving the raw Content
+            // proto bytes that the existing envelope pipeline already decodes
+            // into ContentBody::DecryptionErrorMessage via Content::from_proto.
+            let contents = usmc.contents()?;
+            if contents.first() != Some(&0xC0) {
+                return Err(SignalProtocolError::InvalidMessage(
+                    CiphertextMessageType::Plaintext,
+                    "sealed sender PlaintextContent missing identifier byte",
+                ));
+            }
+            contents[1..].to_vec()
         },
     };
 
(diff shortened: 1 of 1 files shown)


rubdos:
Thank you for bringing this to our attention. The core issue is valid: sealed sender envelopes containing `DecryptionErrorMessage` wrapped in `PlaintextContent` were being dropped and not handled. This should probably have been part of https://github.com/whisperfish/libsignal-service-rs/pull/423

Your implementation approach (manually stripping the 0xC0 marker in libsignal-service-rs) works, but it's not ideal because it duplicates logic: libsignal's Rust code already has `PlaintextContent::try_from()` and `body()` methods that handle this wire format parsing correctly. The test itself is fragile: it tests implementation rather than behaviour.

I do not generally mind the use of agentic coding, but dumping very verbose issue reports, PR descriptions and patch sets wastes time of maintainers. I'm kinda baffled that your Opus 4 didn't catch the better approach here.

Closing in favour of #429

PR #431, opened 2026-05-10
Add with_max_outstanding_keepalives opt-in constructor

tunnell:
`SignalWebSocketProcess::run`'s KA-timer arm closes the WS the
moment ANY keepalive is outstanding (`!is_empty()`), which is
racy when `futures::select!` picks the timer arm before draining
a queued KA response (the macro documents arm dispatch as
"random," not fair). On x86_64 desktop the race is rare enough
to never trip; on a single-threaded async executor on slow
hardware (Precursor, the upcoming Baochip ASIC, possibly the
Whisperfish phone targets) it fires within seconds of opening
every WS.

Adds `SignalWebSocket::with_max_outstanding_keepalives(...,
max: usize)` as an opt-in second constructor; the existing
`pub fn new()` keeps its signature and now delegates with
`DEFAULT_MAX_OUTSTANDING_KEEPALIVES = 1`, preserving today's
behavior exactly. Internally `SignalWebSocketProcess` gains one
`usize` field; the close trigger becomes
`len() >= self.max_outstanding_keepalives`.

**Zero behavior change** for any existing caller. Callers that
need the tolerance (we use `max=3` in `xous-app-signal`)
opt in via the new constructor.

Verified locally: `cargo fmt --check` + `cargo clippy
--all-targets -- -D warnings` clean. Have not independently
MSRV-checked against 1.89, but the patch only uses basic struct
and method patterns; CI should confirm.

Happy to discuss:
- the default value (kept at 1 to preserve current behavior; could
  reasonably be 2 to forgive a single race without much slowdown)
- whether to grow this into a full `SignalWebSocketBuilder` —
  the new fn would become `Builder::build()`'s internal call
- adding a unit test if you want one (the race itself isn't
  trivially testable, but the constructor split is)

Filing as draft to leave room for that discussion before merge.

This work was the result of debugging with a coding agent, where the above findings are the result of this effort.

### DIFF
--- a/src/websocket/mod.rs
+++ b/src/websocket/mod.rs
@@ -91,9 +91,22 @@ struct SignalWebSocketInner {
     stream: Option<SignalRequestStream>,
 }
 
+/// Default value for [`SignalWebSocket::with_max_outstanding_keepalives`].
+/// `1` matches the historical "close on first unacked keepalive"
+/// behavior — every existing caller of [`SignalWebSocket::new`] sees
+/// this default and is therefore unaffected by the introduction of
+/// the knob.
+pub const DEFAULT_MAX_OUTSTANDING_KEEPALIVES: usize = 1;
+
 struct SignalWebSocketProcess {
     /// Whether to enable keep-alive or not (and send a request to this path)
     keep_alive_path: String,
+
+    /// Number of unacked keepalive requests that may be in flight before
+    /// `run()` decides the WebSocket is dead and closes it. Set via
+    /// [`SignalWebSocket::with_max_outstanding_keepalives`]; defaults to
+    /// [`DEFAULT_MAX_OUTSTANDING_KEEPALIVES`].
+    max_outstanding_keepalives: usize,
 
     /// Receives requests from the application, which we forward to Signal.
     requests: mpsc::Receiver<(
@@ -229,8 +242,17 @@ impl SignalWebSocketProcess {
             futures::select! {
                 _ = ka_interval.tick().fuse() => {
                     use prost::Message;
-                    if !self.outgoing_keep_alive_set.is_empty() {
-                        tracing::warn!("Websocket will be closed due to failed keepalives.");
+                    // Close trigger gated by max_outstanding_keepalives,
+                    // which defaults to 1 (the historical "close on any
+                    // unacked keepalive" behavior). Callers that need
+                    // tolerance for executor-scheduling races can pass
+                    // a larger value via with_max_outstanding_keepalives.
+                    if self.outgoing_keep_alive_set.len() >= self.max_outstanding_keepalives {
+                        tracing::warn!(
+                            outstanding = self.outgoing_keep_alive_set.len(),
+                            threshold = self.max_outstanding_keepalives,
+                            "Websocket will be closed due to failed keepalives.",
+                        );
                         if let Err(e) = self.ws.close(reqwest_websocket::CloseCode::Away, None).await {
                             tracing::debug!("Could not close WebSocket: {:?}", e);
                         }
@@ -357,6 +379,32 @@ impl<C: WebSocketType> SignalWebSocket<C> {
         keep_alive_path: String,
         unidentified_push_service: PushService,
     ) -> (Self, impl Future<Output = ()>) {
+        Self::with_max_outstanding_keepalives(
+            ws,
+            keep_alive_path,
+            unidentified_push_service,
+            DEFAULT_MAX_OUTSTANDING_KEEPALIVES,
+        )
+    }
+
+    /// Like [`Self::new`] but lets the caller raise the threshold for
+    /// "outstanding keepalives at which `run()` decides the WebSocket
+    /// is dead and closes it." Default (used by [`Self::new`]) is 1,
+    /// matching the historical close-on-first-unacked-keepalive
+    /// behavior. Callers running on platforms where the
+    /// `futures::select!` arm-dispatch race can spuriously trip the
+    /// close (single-threaded executors, slow rv32 cores, etc.) can
+    /// pass a small constant like 3 to absorb a few cycles of
+    /// scheduling jitter before deciding the connection is dead.
+    /// Real TCP failures still surface independently via the
+    /// underlying transport's `io::Error` path; this only changes the
+    /// application-level KA watchdog.
+    pub fn with_max_outstanding_keepalives(
+        ws: WebSocket,
+        keep_alive_path: String,
+        unidentified_push_service: PushService,
+        max_outstanding_keepalives: usize,
+    ) -> (Self, impl Future<Output = ()>) {
         // Create process
         let (incoming_request_sink, incoming_request_stream) =
             mpsc::unbounded();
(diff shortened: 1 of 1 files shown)


src/websocket/mod.rs
```diff
@@ -91,10 +91,23 @@ struct SignalWebSocketInner {
     stream: Option<SignalRequestStream>,
 }
 
+/// Default value for [`SignalWebSocket::with_max_outstanding_keepalives`].
+/// `1` matches the historical "close on first unacked keepalive"
+/// behavior — every existing caller of [`SignalWebSocket::new`] sees
+/// this default and is therefore unaffected by the introduction of
+/// the knob.
```
rubdos: History is documented by git, not in comments. LLMs have the tendency of spamming code with plan markers and history, which us humans then have to digest. This needs some clean-up. Keep it to one or two lines.

src/websocket/mod.rs
```diff
@@ -91,10 +91,23 @@ struct SignalWebSocketInner {
...
+
 struct SignalWebSocketProcess {
     /// Whether to enable keep-alive or not (and send a request to this path)
     keep_alive_path: String,
 
+    /// Number of unacked keepalive requests that may be in flight before
+    /// `run()` decides the WebSocket is dead and closes it. Set via
+    /// [`SignalWebSocket::with_max_outstanding_keepalives`]; defaults to
+    /// [`DEFAULT_MAX_OUTSTANDING_KEEPALIVES`].
+    max_outstanding_keepalives: usize,
```
rubdos: Either this field, the `const`, or the constructor should be the single-source-of-truth regarding documentation. It's currently spread over many different places which could just link to a single function. Docs are only as good as their maintenance and hence their maintainability.

src/websocket/mod.rs
```diff
@@ -229,8 +242,17 @@ impl SignalWebSocketProcess {
             futures::select! {
                 _ = ka_interval.tick().fuse() => {
                     use prost::Message;
-                    if !self.outgoing_keep_alive_set.is_empty() {
-                        tracing::warn!("Websocket will be closed due to failed keepalives.");
+                    // Close trigger gated by max_outstanding_keepalives,
+                    // which defaults to 1 (the historical "close on any
+                    // unacked keepalive" behavior). Callers that need
+                    // tolerance for executor-scheduling races can pass
+                    // a larger value via with_max_outstanding_keepalives.
```
rubdos: This is highly redundant.


```suggestion
                    // Close if KA's unacked are over max_outstanding_keepalives.
                    // See XXX (the field/constructor/const)
```

src/websocket/mod.rs
```diff
@@ -356,6 +378,32 @@ impl<C: WebSocketType> SignalWebSocket<C> {
...
+    /// matching the historical close-on-first-unacked-keepalive
+    /// behavior. Callers running on platforms where the
+    /// `futures::select!` arm-dispatch race can spuriously trip the
+    /// close (single-threaded executors, slow rv32 cores, etc.) can
+    /// pass a small constant like 3 to absorb a few cycles of
+    /// scheduling jitter before deciding the connection is dead.
+    /// Real TCP failures still surface independently via the
+    /// underlying transport's `io::Error` path; this only changes the
+    /// application-level KA watchdog.
+    pub fn with_max_outstanding_keepalives(
```
rubdos: Again, way too much LLM bloat in the docs which nobody will ever want to read. Additionally, it could be good to add a comment to `new()` to show it's an alias for this new method with the extra default param.

```suggestion
    /// Construct `SignalWebSocket` with configurable KA threshold.
    ///
    ///  A more lenient `max_outstanding_keepalives` can potentially
    /// compensate for spotty or slow networks, or for slower processors.
    pub fn with_max_outstanding_keepalives(
```

rubdos (changes requested):
This sounds more like a network race/spotty or slow network than an actual executor race to me, in which case it could be interesting to try some biased `select!` call (incoming -> outgoing -> KA).

That said, I'm not against having this, although I fear for the potential flooding effect of sending too many KAs at the same time (if network is spotty and there are multiple in-flight). It might be useful to e.g. postpone the next KA send if the previous is still on its way. The runaway potential with this is quite significant...

src/websocket/mod.rs
tunnell: Will do, thanks for flagging this.

src/websocket/mod.rs
tunnell: Good point, I'll take a look

tunnell:
> This sounds more like a network race/spotty or slow network than an actual executor race to me, in which case it could be interesting to try some biased `select!` call (incoming -> outgoing -> KA).
> 
> That said, I'm not against having this, although I fear for the potential flooding effect of sending too many KAs at the same time (if network is spotty and there are multiple in-flight). It might be useful to e.g. postpone the next KA send if the previous is still on its way. The runaway potential with this is quite significant...

It's a slow device since the Precursor is a System-on-Chip running in an FPGA so everything goes pretty slow.  This change got everything working again on the hardware side, but I am mainly making this PR as an upstream commit to avoid having our own branch.  (https://www.crowdsupply.com/sutajio-kosagi/precursor)

Before merging, I'll try your suggestion to see if that helps or if there is an underlying issue, especially if you think my framing is wrong.  Any tips on where the problems might be for a very slow device running your libsignal-service-rs are welcome.

In general, I wanted to propose a change that could be disabled for everybody not running the Xous microkernel on a small embedded system.

tunnell:
Cleaned up the LLM noise (appreciate flagging this).  Let me try select_biased then get back to this, thanks for tip.

rubdos:
> Any tips on where the problems might be for a very slow device running your libsignal-service-rs are welcome.

We have quite some tracing infra in here, and i've used it in the past to try to get some data. Maybe worth checking out something like https://docs.rs/tracing-timing/latest/tracing_timing/ to figure out what is consuming so much CPU. Reordering with a biased select sounds like it would resolve this though, although I fear maybe a consumer of the API could more plausibly cause a deadlock. We'll see.

By the way, I'm generally happy to accomodate to such use cases. I recall having a conversation about using this on the Precursor some years ago. Back then, we had lss split over multiple backends, and I actually intended to have an Embassy-based backend too...

tunnell:
Thanks @rubdos!  If it's fine with you, I'll leave this as a draft for a few weeks to see if it makes more sense to rework this PR.  Are you fine with PRs post-review going back to the drawing board getting converted into a draft? 

I just got over zealous on making sure any fixes I made didn't live just in my own repo, but tried to 'pay it forward'.  My assumption was just that the CPU (it's actually an FPGA emulating a chip) was just super slow.  Along the way, I hit various kernel networking bugs and it has been hard to distinguish sources of errors.  The network stack in this microkernel is stabilizing (other upstream PRs) so I can better assess once it's stable what upstream request makes sense.

I appreciate the input and am just trying to think through the comment about the runaway potential.  And if this fix is the best design for a fix.  

Also, Embassy looks neat.  I hadn't seen it before!

rubdos:
Whichever feels most convenient for you, no worries! I know that developing a whole stack like that, is quite a challenge indeed. :-)

I'm still not sure whether there's an *actual* runaway potential...

tunnell:
I'm going to close this for now.  There are many networking issues in Xous demanding bigger changes there, after which I'll refactor the Signal client to see if this is still needed.  Since things are very 'alpha' on our side, I'll avoid polluting another repo.  Thanks for teaching me also how to make more coherent PRs with these tools, sometimes it's easy to run away with new tech...  I'll swing back in a few months (and potentially re-open).

rubdos:
@tunnell Sure, thanks for keeping in touch! If you find anything new, let us know. Wishing all the best on Xous :-)

PR #432, opened 2026-05-13
feat: Signal Storage Service client + AEP-derived master key

calibrae:
## Summary

Adds the missing piece that modern Signal clients need to do multi-device
contact synchronisation. Modern Signal (2019+) replaced the legacy
`SyncMessage::Contacts` envelope with the **Storage Service** — encrypted,
server-stored, multi-device-shared account state where contacts, groups,
account settings etc. now live. Linked devices that only speak the legacy
sync receive an empty stub from iPhone/Android primaries and end up with
no contacts.

This PR implements the read side of the Storage Service protocol in
`libsignal-service-rs`, plus the `ProvisionMessage` plumbing that
consumers (presage etc.) need to derive the right master key from the
linking handshake.

A companion PR against `whisperfish/presage` adds `Manager::sync_storage()`
on top of this layer — see #__TBD__.

### 1. `protobuf/StorageService.proto` (new, 380 lines)

Vendored from `signalapp/Signal-Android` at
`lib/libsignal-service/src/main/protowire/StorageService.proto`. Defines
the manifest / item / record types. Auto-picked up by the existing
`build.rs` since it lives next to the other protos in `protobuf/`.

### 2. `src/storage_service.rs` (new, ~350 lines)

* **5-endpoint API client** as methods on `PushService`:
  - `get_storage_auth()` → short-lived basic-auth credentials
  - `get_storage_manifest(auth)` → encrypted manifest
  - `get_storage_manifest_if_different_version(auth, version)` → conditional fetch (204 if same)
  - `read_storage_items(auth, keys)` → encrypted records by key

* **Crypto helpers** (matching Signal-Desktop/Android byte-for-byte):
  - `decrypt_storage_manifest(&manifest, &storage_key) -> ManifestRecord`
  - `decrypt_storage_item(&item, &storage_key, record_ikm) -> StorageRecord`
  - AES-256-GCM, 12-byte IV prepended (wire: `iv || ct+tag`)
  - Manifest key: `HMAC-SHA256(storage_key, "Manifest_${version}")`
  - Item key (modern): `HKDF(ikm=recordIkm, info="20240801_SIGNAL_STORAGE_SERVICE_ITEM_" || raw_id, 32)`
  - Item key (legacy fallback): `HMAC-SHA256(storage_key, "Item_" + base64(raw_id))`

* **4 unit tests** covering crypto round-trip + key derivation distinctness.

### 3. `src/provisioning/mod.rs`

`NewDeviceRegistration` now carries two new optional fields read from the
`ProvisionMessage`:

* `master_key: Option<Vec<u8>>` — deprecated `masterKey` proto field 13.
  Comment in the proto says it's "Deprecated, but required by linked
  devices", so iPhones still send it. Older clients use this directly.
* `account_entropy_pool: Option<String>` — modern `accountEntropyPool`
  proto field 15. The 64-char alphanumeric Root of All Entropy.
  Consumers should derive the canonical MasterKey from this via
  `libsignal_account_keys::AccountEntropyPool::derive_svr_key()` and
  prefer it over the deprecated field.

Both fields default to `None` if the primary doesn't send them, so
this is fully backwards-compatible with anything that ignored the
proto fields before.

### 4. `Cargo.toml` + `src/lib.rs`

* Adds `libsignal-account-keys` as a direct dep (same git tag as the
  rest of the libsignal family).
* `pub use libsignal_account_keys;` at the crate root so consumers can
  reach `AccountEntropyPool` without taking a separate dep.

## Why now / why this scope

Modern Signal-iOS and Signal-Android both treat the legacy
`SyncMessage::Contacts` blob as essentially-removed — they ship a stub
on receipt of a `SyncMessage::Request::Contacts` so that very-old
clients don't crash, but the real contact list has moved to Storage
Service. The Whisperfish issue tracker has long had requests for
Storage Service support
([whisperfish/whisperfish#321](https://github.com/whisperfish/whisperfish/issues/321),
the issue is years old). This PR is the lower half of that work.

The write side (`PUT /v1/storage`, conflict resolution, item insertion
+ deletion) is intentionally left out — it requires implementing the
full version-conflict dance (HTTP 409 → server returns new manifest →
client diff-applies → retry). Read-only is enough to make linked
secondaries usable; writes can be added later as a follow-up by anyone
who needs to edit the contact list from a libsignal-service consumer.

## Reference

* Canonical implementation: `signalapp/Signal-Android`
  - API client: `lib/libsignal-service/src/main/java/org/whispersystems/signalservice/api/storage/StorageServiceApi.kt`
  - Crypto: `core/models-jvm/src/main/java/org/signal/core/models/storageservice/SignalStorageCipher.kt`
  - Key derivation: `core/models-jvm/src/main/java/org/signal/core/models/storageservice/StorageKey.kt`
  - RecordIkm: `lib/libsignal-service/src/main/java/org/whispersystems/signalservice/api/storage/RecordIkm.kt`
* Signal-Desktop parallel: `ts/Crypto.node.ts` (`deriveStorageManifestKey`, `deriveStorageItemKey`, `decryptProfile`)

## Tested end-to-end

Built against a Signal-iOS 7.x primary device on a freshly-linked
secondary running `presage` + the companion storage-service branch:
**165/165 contacts pulled from Storage Service on first link**, all
decrypted cleanly, all `ContactRecord` fields parsed (aci/aci_binary,
e164, profile_key, given/family/nickname/system_*).

Pairing the storage_service_key derivation against the existing
`StorageServiceKey::from_master_key` test in `master_key.rs` confirmed
the master key path is the same one Signal-Desktop uses (verified
against the documented v7.23.0 reference vector still in that file).

## CI

* `cargo fmt --check` — clean
* `cargo build --all-features` — clean
* `cargo build --no-default-features --features phonenumber` — clean
  (the configuration `presage-store-sqlite` uses)
* `cargo test --lib storage_service` — 4/4 pass
* MSRV 1.89 — `derive_svr_key()` from `libsignal-account-keys` is
  stable on that, no nightly/recent features used

## Backwards compatibility

* Adding optional fields to `NewDeviceRegistration` is non-breaking
  for direct consumers (struct lookups by name still work; struct
  literals would need to be updated but `#[non_exhaustive]` would
  protect against that — happy to add it if you want).
* All other additions are net-new modules / re-exports; no existing
  API touched.

## Open to feedback on

* Naming of `decrypt_storage_manifest` / `decrypt_storage_item` —
  could also be `StorageManifest::decrypt(self, key)` constructor
  methods if you prefer the OO style.
* Whether `libsignal-account-keys` should be re-exported under
  `protocol::` instead of crate root (kept at root because it isn't
  technically part of libsignal-protocol, but happy to move).
* Whether to gate `storage_service` behind a feature flag. Currently
  unconditional — the module is ~350 LOC and the only new transitive
  dep is `libsignal-account-keys` which is small.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/protobuf/StorageService.proto
+++ b/protobuf/StorageService.proto
@@ -1,0 +1,379 @@
+/**
+ * Copyright (C) 2019 Open Whisper Systems
+ *
+ * Licensed according to the LICENSE file in this repository.
+ */
+syntax = "proto3";
+
+package signalservice;
+
+option java_package         = "org.whispersystems.signalservice.internal.storage.protos";
+option java_multiple_files  = true;
+
+enum OptionalBool {
+  UNSET    = 0;
+  ENABLED  = 1;
+  DISABLED = 2;
+}
+
+message StorageManifest {
+  uint64 version = 1;
+  bytes  value   = 2;
+}
+
+message StorageItem {
+  bytes key   = 1;
+  bytes value = 2;
+}
+
+message StorageItems {
+  repeated StorageItem items = 1;
+}
+
+message ReadOperation {
+  repeated bytes readKey = 1;
+}
+
+message WriteOperation {
+           StorageManifest manifest   = 1;
+  repeated StorageItem     insertItem = 2;
+  repeated bytes           deleteKey  = 3;
+           bool            clearAll   = 4;
+}
+
+message ManifestRecord {
+  message Identifier {
+    enum Type {
+      UNKNOWN                 = 0;
+      CONTACT                 = 1;
+      GROUPV1                 = 2;
+      GROUPV2                 = 3;
+      ACCOUNT                 = 4;
+      STORY_DISTRIBUTION_LIST = 5;
+      CALL_LINK               = 7;
+      CHAT_FOLDER             = 8;
+      NOTIFICATION_PROFILE    = 9;
+    }
+
+    bytes raw  = 1;
+    Type  type = 2;
+  }
+
+           uint64     version      = 1;
+           uint32     sourceDevice = 3;
+  repeated Identifier identifiers  = 2;
+           bytes      recordIkm    = 4;
+  // Next ID: 5
+}
+
+message StorageRecord {
+  oneof record {
+    ContactRecord               contact               = 1;
+    GroupV1Record               groupV1               = 2;
+    GroupV2Record               groupV2               = 3;
+    AccountRecord               account               = 4;
+    StoryDistributionListRecord storyDistributionList = 5;
+    CallLinkRecord              callLink              = 7;
+    ChatFolderRecord            chatFolder            = 8;
+    NotificationProfile         notificationProfile   = 9;
+  }
+}
+
+
+// If unset - computed as the value of the first byte of SHA-256(msg=CONTACT_ID)
+// modulo the count of colors. Once set the avatar color for a recipient is
+// never recomputed or changed.
+//
+// `CONTACT_ID` is the first available identifier from the list:
+// - ServiceIdToBinary(ACI)
+// - E164
+// - ServiceIdToBinary(PNI)
+// - Group Id
+enum AvatarColor {
+  A100 = 0;
+  A110 = 1;
+  A120 = 2;
+  A130 = 3;
+  A140 = 4;
+  A150 = 5;
+  A160 = 6;
+  A170 = 7;
+  A180 = 8;
+  A190 = 9;
+  A200 = 10;
+  A210 = 11;
+}
+
+message ContactRecord {
+  enum IdentityState {
+    DEFAULT    = 0;
+    VERIFIED   = 1;
+    UNVERIFIED = 2;
+  }
+
+  message Name {
+    string given  = 1;
+    string family = 2;
+  }
+
+  string        aci                     = 1;
+  string        e164                    = 2;
+  string        pni                     = 15;
+  bytes         profileKey              = 3;
+  bytes         identityKey             = 4;
+  IdentityState identityState           = 5;
+  string        givenName               = 6;
+  string        familyName              = 7;
+  string        username                = 8;
+  bool          blocked                 = 9;
+  bool          whitelisted             = 10;
+  bool          archived                = 11;
+  bool          markedUnread            = 12;
+  uint64        mutedUntilTimestamp     = 13;
+  bool          hideStory               = 14;
+  uint64        unregisteredAtTimestamp = 16;
+  string        systemGivenName         = 17;
+  string        systemFamilyName        = 18;
+  string        systemNickname          = 19;
+  bool          hidden                  = 20;
+  bool          pniSignatureVerified    = 21;
+  Name          nickname                = 22;
+  string        note                    = 23;
+  optional AvatarColor avatarColor      = 24;
+  bytes         aciBinary               = 25; // 16-byte UUID
+  bytes         pniBinary               = 26; // 16-byte UUID
+  // Next ID: 27
+}
+
+message GroupV1Record {
+  bytes  id                  = 1;
+  bool   blocked             = 2;
+  bool   whitelisted         = 3;
+  bool   archived            = 4;
+  bool   markedUnread        = 5;
+  uint64 mutedUntilTimestamp = 6;
+}
+
+message GroupV2Record {
+  enum StorySendMode {
+    DEFAULT  = 0;
+    DISABLED = 1;
+    ENABLED  = 2;
+  }
+
+  bytes         masterKey                    = 1;
+  bool          blocked                      = 2;
+  bool          whitelisted                  = 3;
+  bool          archived                     = 4;
+  bool          markedUnread                 = 5;
+  uint64        mutedUntilTimestamp          = 6;
+  bool          dontNotifyForMentionsIfMuted = 7;
+  bool          hideStory                    = 8;
+  reserved   /* storySendEnabled */            9;
+  StorySendMode storySendMode                = 10;
+  optional AvatarColor avatarColor           = 11;
+}
+
+message Payments {
+  bool  enabled = 1;
+  bytes entropy = 2;
+}
+
+message AccountRecord {
+
+  enum PhoneNumberSharingMode {
+    UNKNOWN   = 0;
+    EVERYBODY = 1;
+    NOBODY    = 2;
+  }
+
+  message PinnedConversation {
+    message Contact {
+      string serviceId       = 1;
+      string e164            = 2;
+      bytes  serviceIdBinary = 3; // service ID binary (i.e. 16 byte UUID for ACI, 1 byte prefix + 16 byte UUID for PNI)
+    }
+
+    oneof identifier {
+      Contact contact        = 1;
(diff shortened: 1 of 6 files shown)


protobuf/StorageService.proto
```diff
@@ -0,0 +1,380 @@
+/**
+ * Copyright (C) 2019 Open Whisper Systems
+ *
+ * Licensed according to the LICENSE file in this repository.
+ */
```
rubdos, line 5: We have an import script in `/prototbuf/`, make sure that we have the correct references there.

src/provisioning/mod.rs
```diff
@@ -228,6 +241,11 @@ pub async fn link_device<
             .profile_key
             .ok_or(ProvisioningError::MissingProfileKey)?;
 
+        // Optional in the proto (field 13 deprecated, field 15 modern).
+        // We accept either — presage will prefer AEP if present.
+        let master_key = message.master_key.clone();
+        let account_entropy_pool = message.account_entropy_pool.clone();
+
```
rubdos, line 249: Can't you just move out of the message and avoid the clone?

src/storage_service.rs
```diff
@@ -0,0 +1,369 @@
...
+
+impl PushService {
+    /// Fetch short-lived basic-auth credentials for the storage service.
+    ///
+    /// `GET /v1/storage/auth`
+    ///
+    /// Pass the returned [`HttpAuth`] to the other storage methods. The
+    /// credentials are valid for ~24h on the server side; in practice fetch
+    /// a fresh pair at the start of every sync.
+    pub async fn get_storage_auth(&self) -> Result<HttpAuth, ServiceError> {
```
rubdos: I would prefer a strong type around HttpAuth, which then combined with a PushService can provide the remaining methods. See e.g. `AccountManager`, you could have a StorageService or so which encapsulates a PushService and the auth.

That way, we can also keep the network logic separate from the API.

src/storage_service.rs
```diff
@@ -0,0 +1,369 @@
...
+        plaintext.encode(&mut pt_buf).unwrap();
+
+        // Manual encrypt to produce the wire blob.
+        let cipher = Aes256Gcm::new(Key::<Aes256Gcm>::from_slice(&key));
+        let nonce_bytes = [0u8; IV_LEN];
+        let ct = cipher
+            .encrypt(Nonce::from_slice(&nonce_bytes), pt_buf.as_slice())
+            .unwrap();
+        let mut blob = nonce_bytes.to_vec();
+        blob.extend_from_slice(&ct);
```
rubdos: At this point we could as well just provide the encryption methods as well, for when the other direction gets implemented. That simplifies your test, makes it look less silly, and provides some more value.

src/storage_service.rs
```diff
@@ -0,0 +1,369 @@
...
+        .service_error_for_status()
+        .await?
+        .protobuf()
+        .await
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Crypto
+// ---------------------------------------------------------------------------
```
rubdos: If you build the "manager" or "service" struct, these will all naturally move into that bespoke module, and become private methods in that struct.

I would also appreciate a filepath to the Android source code or whatever you used, that makes it way easier to trace for the future and to review.

rubdos (commented):
Thanks for your contribution. The code looks quite alright, although it doesn't seem to follow the regular architecture that we tend to use. I'm quite sure that one more pass through Claude will clean this up quite nicely.

Your PR description is immensely verbose, and it wastes reviewer's time. You could've just written something like this:

> Signal dropped `SyncMessage::Contacts` in favor of Storage Service. Linked devices without this support get an empty contact list.

src/storage_service.rs
```diff
@@ -0,0 +1,369 @@
...
+}
+
+/// Modern item key (with `recordIkm`):
+/// `HKDF-SHA256(ikm = recordIkm, info = "20240801_SIGNAL_STORAGE_SERVICE_ITEM_" || raw_id, len = 32)`.
+fn derive_item_key_from_ikm(record_ikm: &[u8], raw_id: &[u8]) -> [u8; 32] {
+    let mut info = b"20240801_SIGNAL_STORAGE_SERVICE_ITEM_".to_vec();
+    info.extend_from_slice(raw_id);
+    let hk = Hkdf::<Sha256>::new(None, record_ikm);
+    let mut okm = [0u8; 32];
+    hk.expand(&info, &mut okm)
```
rubdos: Use [expand_multi_info](https://docs.rs/hkdf/latest/hkdf/type.Hkdf.html#method.expand_multi_info) to avoid allocation.

```suggestion
    hk.expand_multi_info(&[info, raw_id], &mut okm)
```

src/storage_service.rs
```diff
@@ -0,0 +1,369 @@
...
+
+/// Legacy item key (no `recordIkm`):
+/// `HMAC-SHA256(storage_key, "Item_" + base64(raw_id))`.
+fn derive_item_key_legacy(
+    storage_key: &StorageServiceKey,
+    raw_id: &[u8],
+) -> [u8; 32] {
+    let b64 = base64::engine::general_purpose::STANDARD.encode(raw_id);
+    let info = format!("Item_{}", b64);
+    hmac_sha256(&storage_key.inner, info.as_bytes())
```
rubdos: If you use the Hmac construction inline instead of with the wrapper `hmac_sha256`, you can just call `update` multiple times and avoid the second allocation. Same above.

src/storage_service.rs
```diff
@@ -0,0 +1,369 @@
...
+    blob: &[u8],
+) -> Result<Vec<u8>, StorageServiceError> {
+    if blob.len() < IV_LEN {
+        return Err(StorageServiceError::BlobTooShort);
+    }
+    let (iv, ct) = blob.split_at(IV_LEN);
+    let cipher = Aes256Gcm::new(Key::<Aes256Gcm>::from_slice(key));
+    cipher
+        .decrypt(Nonce::from_slice(iv), ct)
+        .map_err(|_| StorageServiceError::Decrypt)
```
rubdos: I would collapse both error types onto "Invalid" or so. There's no reason to distinguish them from the user-facing side.

rubdos (changes requested):
Some more comments for the crypto impl. After refactor, I'll give that another pass.

calibrae:
Refactored — thanks for the steer.

- `StorageService` struct now wraps `PushService` + storage auth + the key, mirroring `AccountManager`. API methods + crypto are its methods; callers get decrypted protobufs out.
- Added `encrypt_*` alongside `decrypt_*` — round-trip tests use them instead of hand-rolled AES-GCM.
- `StorageService.proto` registered in `update-protos.sh`, re-vendored from Signal-Android v8.3.1.
- `provisioning/mod.rs`: moved out of `message` instead of cloning.
- `expand_multi_info` for the HKDF path; inline HMAC `update()` calls for the legacy/manifest keys — no intermediate allocs.
- Error enum collapsed to `Invalid` + `Service`.
- Android source paths in the module doc.

fmt + clippy + build (all features) + tests green. Ready for another pass.

rubdos:
I've made some more changes on top of your code. Merged!

PR #435, opened 2026-05-14
chore: update to libsignal v0.94.0

rubdos:
- Update all libsignal crates from v0.91.0 to v0.94.0
- Add local_address parameter to message_decrypt_signal calls
- Add local_address parameter to process_prekey_bundle calls
- Make ExpiringProfileKeyCredentialPresentation version a const generic parameter on create_member_presentation with explicit specification at all call sites using PRESENTATION_VERSION_3


rubdos:
cc @roder

PR #436, opened 2026-05-14
Key transparency

rubdos:
This is currently very much slop.

The goal is to test this with the example that's currently shipped. You can easily test this against your own key (which we should probably actually do in our apps):

```sh
cargo run --example key_transparency --features key-transparency -- --password $PASSWORD --aci $ACI --target-aci $ACI --target-identity $IDENTITY
```


rubdos:
https://github.com/signalapp/Signal-Server/commit/1b09529ecefc7a69127d5f624458259ac53917fa

... guess I'll have to start over.

PR #438, opened 2026-05-23
fix: contact sync to handle aci_binary field.

### LINKED ISSUE #437: Contact sync broken: `Contact::from_proto` only reads `aci` string field, missing `aci_binary`
After updating to the latest libsignal-service-rs (`bd585ac`) which added `aci_binary` (field 13) to the `ContactDetails` proto, contact sync from the primary device silently drops all contacts with a `MissingUuid` error.

When linking as a secondary device, the primary device sends a `SyncMessage` with `contacts` containing a CDN blob. The blob downloads successfully (we see the HTTP 200 from CDN), but after decryption, **every single `ContactDetails` entry** fails `prost::Message::decode_length_delimited` → `Contact::from_proto` with:

```
ParseContactError::MissingUuid
```

`from_proto` only reads `contact_details.aci` (the string UUID field, field 9). It does not check `contact_details.aci_binary` (the 16-byte binary UUID, field 13). It seems the primary device (latest Signal Android) now sends `aci_binary` and omits or doesn't populate `aci` in the blob format.

→  Contacts sync appears to succeed (no errors surfaced to the caller — `retrieve_contacts` returns an iterator of `Result`s, and `filter_map(Result::ok)` silently drops all failures), but zero contacts are persisted.

**Workaround**→ A local patch to `Contact::from_proto` that falls back to `aci_binary` when `aci` is absent resolves the issue.

Is this a known change in the primary device's contact sync format? Is `from_proto` expected to handle both fields?

rubdos:
Fixes #437

@hollisticated-horse, would be good if you could confirm this works. I haven't tested any contact sync since forever.

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -48,11 +48,11 @@ impl Contact {
         avatar_data: Option<Bytes>,
     ) -> Result<Self, ParseContactError> {
         Ok(Self {
-            uuid: contact_details
-                .aci
-                .as_ref()
-                .ok_or(ParseContactError::MissingUuid)?
-                .parse()?,
+            uuid: crate::utils::parse_uuid_with_fallback(
+                contact_details.aci_binary.as_deref(),
+                contact_details.aci.as_deref(),
+            )
+            .ok_or(ParseContactError::MissingUuid)?,
             phone_number: contact_details.number.as_ref().and_then(|n| {
                 n.parse::<E164>()
                     .inspect_err(|e| {


hollisticated-horse:
I believe that fixed it for me !

rubdos:
Override clippy rule because unrelated to this PR. I'll probably find some time later next week for some cleanups. Cheers-!

PR #439, opened 2026-05-24
Primary registration fix: small fixes

rubdos:
These are discrepancies found with Mistral analyzing the server and lss code bases. There's probably more, but this should clean up some stuff already.

Incorporate verification code submission endpoint field discrepancies by updating libsignal-service-rs to match Signal-Server's VerificationSessionResponse, and enable response body exposure for debugging 400 errors.

Builds on (and merges into) fix-primary-registration by @gferon


rubdos:
Thought I had more than I actually do... Reopening the error handling against main.

PR #440, opened 2026-05-24
feat: capture body in UnhandledResponseCode

rubdos:
Refactors service_error_for_status to always fall back to raw body processing in cases where the mechanical error handling fails.

Especially useful in debugging 400 Bad Request cases.

### DIFF
--- a/src/websocket/mod.rs
+++ b/src/websocket/mod.rs
@@ -182,14 +182,24 @@ impl SignalWebSocketProcess {
                     } else if let Some(_x) =
                         self.outgoing_keep_alive_set.take(&id)
                     {
-                        let status_code = response.status();
-                        if status_code != 200 {
+                        let status = reqwest::StatusCode::from_u16(
+                            response.status() as _,
+                        )
+                        .map_err(|e| {
+                            ServiceError::IO(std::io::Error::other(format!(
+                                "invalid http status code {} - {e}",
+                                response.status()
+                            )))
+                        })?;
+                        if !status.is_success() {
                             tracing::warn!(
-                                status_code,
-                                "response code for keep-alive != 200"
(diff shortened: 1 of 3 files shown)


rubdos:
@gferon / @direc85 this should make it easier to test WS-based registration.

src/websocket/mod.rs
```diff
@@ -182,14 +182,24 @@ impl SignalWebSocketProcess {
...
-                        let status_code = response.status();
-                        if status_code != 200 {
+                        let status = reqwest::StatusCode::from_u16(
+                            response.status() as _,
+                        )
+                        .map_err(|e| {
+                            ServiceError::IO(std::io::Error::other(format!(
+                                "invalid http status code {} - {e}",
+                                response.status()
+                            )))
```
Schmiddiii, line 192: Why an IO error and not UnhandledResponseCode?

Schmiddiii (approved):
Code-wise, looks good to me.
Not sure if I can test it though.

src/websocket/mod.rs
rubdos: I imagine the UnhandledResponseCode to be a *valid* status code (parsed like below), but unknown/unhandled. This branch is for invalid codes. Correct me if that doesn't make sense :-)

Schmiddiii: I see your interpretation. I still find an IO error to be a weird choice. But as the description of the error is good, I don't think this should lead to confusion.

rubdos: Hmm, maybe `InvalidFrame` is more suited, but that takes a `&'static str`. Either way, if this pops up it a server-side error or a bug in reqwest. I don't think this code path will ever pop up...

PR #442, opened 2026-06-11
feat: envelope ack handle

rubdos:
Very much vibe coded. This should allow consumers of the library to crash and leave the message on the server. Caveat being that if a crash happens during processing *because* of processing, that the crashing will keep happening until the message processing is corrected in a follow-up update.

### DIFF
--- a/src/messagepipe.rs
+++ b/src/messagepipe.rs
@@ -20,10 +20,58 @@ use crate::{
     websocket::{self, SignalWebSocket},
 };
 
+/// Handle for acknowledging an envelope back to the Signal server.
+///
+/// Take this handle out of [`Incoming::Envelope`] and either:
+///
+/// * **Call [`AckHandle::ack`]** to acknowledge immediately - the server
+///   removes the envelope from its queue.  This is useful when you want to
+///   ack before the end of the scope (e.g. after a quick validation).
+/// * **Move the handle into a future or closure** and let it drop later;
+///   on drop it will auto-acknowledge.  If the thread is panicking when
+///   the handle drops, the envelope is *not* acknowledged and will be
+///   re-delivered by the server on the next connection.
+#[derive(Debug)]
+pub struct AckHandle(
+    Option<(
+        oneshot::Sender<WebSocketResponseMessage>,
+        WebSocketResponseMessage,
+    )>,
+);
+
+impl AckHandle {
+    /// Acknowledge this envelope immediately.
+    ///
+    /// Acknowledge the envelope immediately, before the handle would
+    /// otherwise be dropped.  Useful when you want to signal the server
+    /// early (e.g. after a quick validation) and don't need to defer
+    /// the ack to end-of-scope.
+    ///
+    /// Equivalent with `drop(AckHandle)`.
+    pub fn ack(self) {}
+}
+
+impl Drop for AckHandle {
+    fn drop(&mut self) {
+        let Some((responder, response)) = self.0.take() else {
+            return; // ack() was already called
+        };
+        if std::thread::panicking() {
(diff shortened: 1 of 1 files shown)


src/messagepipe.rs
```diff
@@ -20,27 +20,102 @@ use crate::{
...
+    }
+
+    fn send_ack(&mut self) {
+        if let Some(responder) = self.responder.take() {
+            let response = WebSocketResponseMessage {
+                id: self.request_id,
+                status: Some(200),
+                message: Some("OK".to_string()),
+                ..Default::default()
+            };
```
rubdos: I find this more the responsibility of the socket process itself; the responder should probably just consume a marker type `struct Ack;`.

src/messagepipe.rs
```diff
@@ -20,27 +20,102 @@ use crate::{
...
+/// * When constructed with `auto_ack = true` (the default):
+///   - It will automatically acknowledge the envelope, **unless** the thread
+///     is currently panicking, in which case the envelope is left
+///     unacknowledged (the server will re-deliver it).
+/// * When constructed with `auto_ack = false`:
+///   - It will never auto-acknowledge and will emit a warning log instead.
+///   - The envelope is left unacknowledged (the server will re-deliver it).
+#[derive(Debug)]
+pub struct AckHandle {
+    request_id: Option<u64>,
```
rubdos: Does this actually need to be an Option? That doesn't sound logical to me.

src/messagepipe.rs
```diff
@@ -20,27 +20,102 @@ use crate::{
...
+    ///
+    /// When `true` (the default), envelopes are auto-acknowledged if the
+    /// caller drops the handle without calling [`AckHandle::ack`], unless
+    /// the thread is currently panicking.
+    ///
+    /// When `false`, the caller **must** call [`AckHandle::ack`] explicitly;
+    /// a warning is logged if the handle is dropped without acknowledgement.
+    pub fn set_auto_ack(&mut self, auto_ack: bool) {
+        self.auto_ack = auto_ack;
+    }
```
rubdos: I'm not sure whether we'll want to turn this off in any app, since I can't imagine a situation where we want to not-ack that is not a panic. Maybe this is an unnecessary option.

src/messagepipe.rs
```diff
@@ -67,42 +142,63 @@ impl MessagePipe {
...
+                    id: request.id,
+                    status: Some(400),
+                    message: Some("Bad Request".to_string()),
+                    ..Default::default()
+                };
+                let _ = responder.send(response);
                 return Err(ServiceError::InvalidFrame {
                     reason: "request without body.",
                 });
             };
```
rubdos: I feel like this change is unrelated, incorrect, and hallucinated by Deepseek v4 flash. Will reiterate.

src/messagepipe.rs
```diff
@@ -67,42 +142,63 @@ impl MessagePipe {
...
+    ///   response is sent immediately, as before.
+    /// * For **envelope** requests, the ack is deferred into the
+    ///   [`AckHandle`] and the caller must acknowledge it explicitly.
     async fn process_request(
         &mut self,
         request: WebSocketRequestMessage,
         responder: oneshot::Sender<WebSocketResponseMessage>,
     ) -> Result<Option<Incoming>, ServiceError> {
-        // Java: MessagePipe::read
-        let response = WebSocketResponseMessage::from_request(&request);
```
rubdos, line 76: This is the logic that our ack system should also use, instead of manually constructing the ack.

rubdos (commented):
A bit too much change here.

PR #443, opened 2026-06-20
Ignore unusable sender key records when encrypting

Schmiddiii:
Session reset works by marking sessions as archived (they are still in the DB, just marked as unusable). Encrypting for sending did not check whether a sender chain is usable, and therefore threw an error when trying to get the session version. This commit verifies that a sender session is usable before trying to encrypt something with it.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -11,8 +11,8 @@ use libsignal_protocol::{
     PlaintextContent, PreKeySignalMessage, PreKeyStore, ProtocolAddress,
     ProtocolStore, PublicKey, SealedSenderDecryptionResult, SenderCertificate,
     SenderKeyDistributionMessage, SenderKeyStore, ServiceId, SessionStore,
-    SignalMessage, SignalProtocolError, SignedPreKeyStore, Timestamp,
-    UnidentifiedSenderMessageContent,
+    SessionUsabilityRequirements, SignalMessage, SignalProtocolError,
+    SignedPreKeyStore, Timestamp, UnidentifiedSenderMessageContent,
 };
 use prost::Message;
 use rand::{rng, CryptoRng, Rng};
@@ -486,6 +486,16 @@ where
             .ok_or_else(|| {
             SignalProtocolError::SessionNotFound(address.clone())
         })?;
+
+        let record_usable = session_record
+            .has_usable_sender_chain(
+                SystemTime::now(),
+                SessionUsabilityRequirements::NotStale,
+            )
+            .unwrap_or(false);
+        if !record_usable {
+            Err(SignalProtocolError::SessionNotFound(address.clone()))?;
+        }
 
         let padded_content =
             add_padding(session_record.session_version()?, content)?;


Schmiddiii:
CI-failure unrelated to this MR.

Schmiddiii:
Noticed that this is not really needed, I can immediately delete the sessions as well, as that would be done either way on a `SessionNotFound` error. Feel free to close this if you think immediate-delete is better.

Schmiddiii:
Actually, deleting sessions is not correct, see [this comment](https://github.com/whisperfish/presage/pull/415#issuecomment-4762840932); so this is indeed needed again to my knowledge.

src/cipher.rs
```diff
@@ -487,6 +487,14 @@ where
             SignalProtocolError::SessionNotFound(address.clone())
         })?;
 
+        let record_usable = session_record.has_usable_sender_chain(
+            SystemTime::UNIX_EPOCH,
+            SessionUsabilityRequirements::all(),
+        );
+        if record_usable.is_err() || record_usable.is_ok_and(|r| !r) {
```
rubdos:
```suggestion
        if record_usable.map_or(true, |r| !r) {
```

src/cipher.rs
```diff
@@ -487,6 +487,14 @@ where
             SignalProtocolError::SessionNotFound(address.clone())
         })?;
 
+        let record_usable = session_record.has_usable_sender_chain(
+            SystemTime::UNIX_EPOCH,
+            SessionUsabilityRequirements::all(),
```
rubdos:
```suggestion
            SessionUsabilityRequirements::NotStale,
```

`::all()` will filter out any sessions that do not incorporate whatever new fancy features Signal has built in into the Signal protocol. I.e., in the current case, this will force all sessions to be `EstablishedWithPqxdh` *and* `Spqr`, which is not something we currently need to manually enforce.

`::empty()` should only check whether the session is initialized and not emptied, i.e., it's a valid session without any further requirements. `NotStale` additionally checks whether any prekey init is outstanding for too long, and marks it as unusable.

src/cipher.rs
```diff
@@ -487,6 +487,14 @@ where
             SignalProtocolError::SessionNotFound(address.clone())
         })?;
 
+        let record_usable = session_record.has_usable_sender_chain(
+            SystemTime::UNIX_EPOCH,
```
rubdos:
```suggestion
            SystemTime::now(),
```

or whatever the correct syntactic variant is. UNIX_EPOCH is presumably coming from Signal's test code, where test records are made at UNIX_EPOCH for deterministic testing. It doesn't matter though for the intended test, especially if you set `empty()` or `NotStale` as per below.

rubdos (commented):
This is mostly @gferon's terrain, but as far as I understand your concern and change is correct, and what I nitted down here would be a bit correctererer.

src/cipher.rs
rubdos: I believe this misses an `!`, sorry about that.

Schmiddiii: Replaced it with `!record_usable.unwrap_or(false)`, which should be same as far as I can see.

PR #444, opened 2026-06-26
feat: Member::label/label_emoji

rubdos:
(currently testing)


rubdos:
Tested and working. Notably, I'm getting ZK group verification errors on P2P member label group updates currently.

rubdos:
Tested and working even more.

PR #446, opened 2026-06-27
Tolerate plaintext device names

direc85:
Apparently sometimes the device name can be plaintext, possibly not even base64 encoded, so let's not throw errors when that happens. This is visible in Whisperfish when opening linked devices list.

### DIFF
--- a/src/account_manager.rs
+++ b/src/account_manager.rs
@@ -990,8 +990,23 @@ fn decrypt_device_name_from_device_info(
     string: &str,
     aci: &IdentityKeyPair,
 ) -> Result<String, ServiceError> {
-    let data = BASE64_RELAXED.decode(string)?;
-    let name = DeviceName::decode(&*data)?;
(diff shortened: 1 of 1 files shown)


src/account_manager.rs
```diff
@@ -990,8 +990,23 @@ fn decrypt_device_name_from_device_info(
...
+    let Ok(data) = BASE64_RELAXED.decode(string) else {
+        tracing::trace!(
+            "Device name not base64 encoded, assuming plaintext: {}",
+            string.to_string()
+        );
+        return Ok(string.to_string());
+    };
+    let name = match DeviceName::decode(data.as_slice()) {
+        Ok(decoded) => decoded,
+        Err(e) => {
```
rubdos:
```suggestion
        Err(_e) => {
```

direc85: Thanks, I don't have my pre-push scripts working at the moment...

PR #447, opened 2026-06-27
Send AEP to linked device

direc85:
Linking another device requires master key and account entropy pool to be sent, so let's send them. Reordered the fields to proto file order, that makes this messier than it is. EBK and MRBK are still null-safe, apparently, and they require **a lot of** plumbing to get going, so have them as options now.

This raises the amount of parameters to 10 which means we probably should pass a struct instead...

### DIFF
--- a/src/account_manager.rs
+++ b/src/account_manager.rs
@@ -20,7 +20,6 @@ use tracing_futures::Instrument;
 use zkgroup::profiles::ProfileKey;
 
 use crate::content::ContentBody;
-use crate::master_key::MasterKey;
 use crate::pre_keys::{
     KyberPreKeyEntity, PreKeyEntity, PreKeysStore, SignedPreKeyEntity,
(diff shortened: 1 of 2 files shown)


src/account_manager.rs
```diff
@@ -364,6 +365,9 @@ impl AccountManager {
         pni_identity_store: &dyn IdentityKeyStore,
         credentials: ServiceCredentials,
         master_key: Option<MasterKey>,
+        ephemeral_backup_key: Option<[u8; 32]>,
+        account_entropy_pool: AccountEntropyPool,
+        media_root_backup_key: Option<[u8; 32]>,
```
rubdos: I suppose these five could become a struct if you're feeling like it, but I don't mind slapping a `#[allow(clippy` on here.

direc85: Yeap, made it a `ProvisioningSecrets` struct and it's a lot cleaner now.

rubdos (approved):

PR #448, opened 2026-06-28
Username cleanup and example

rubdos:


src/websocket/usernames.rs
```diff
@@ -51,23 +53,31 @@ impl SignalWebSocket<Unidentified> {
...
+    /// `link` is accepted either as a full `https://signal.me/#eu/<payload>`
+    /// link or as the bare `<payload>` that follows `#eu/`. The payload is the
+    /// URL-safe base64 encoding of the 32-byte link entropy followed by the
+    /// 16-byte link handle UUID.
     // Based on libsignal-net
     pub async fn look_up_username_link(
         &mut self,
-        uuid: uuid::Uuid,
-        entropy: &[u8; usernames::constants::USERNAME_LINK_ENTROPY_SIZE],
+        link: &str,
```
Schmiddiii: Nitpick: Maybe only accept `url::Url`? Would prevent bare-payload lookup though, so I would also be fine with keeping it as-is. Same also for everywhere else.

src/websocket/usernames.rs
```diff
@@ -121,3 +131,145 @@ impl SignalWebSocket<Unidentified> {
...
+    ServiceError,
+> {
+    let payload = link
+        .rsplit_once("#eu/")
+        .map(|(_, payload)| payload)
+        .unwrap_or(link);
+
+    let bytes = BASE64_URL_SAFE_NO_PAD
+        .decode(payload)
+        .or_else(|_| BASE64_RELAXED.decode(payload))?;
```
Schmiddiii: Question: Have you actually observed `BASE64_RELAXED`-links or is that just a probably undeeded fallback?

Schmiddiii (approved):

src/websocket/usernames.rs
rubdos: Agree on the link generation site, but indeed accepting the bare payload is maybe a feature we want to retain.

src/websocket/usernames.rs
rubdos: That was an early unneeded fallback. libsignal-net doesn't do this.

Issue #451, opened 2026-07-04
Expose and use /v1/messages/multi_recipient in sender key sending


rubdos:
I'm working on using and distributing sender keys, but for now we're still deliverying to all recipients in a loop. I'm postponing this because the PR is getting large as-is.

PR #453, opened 2026-07-04
Sender keys

rubdos:
Still requires #451 for true efficiency gains, but this implements sender-side sender keys. Tested working, but I still need a few passes through the code to get rid of the "vibes".

### DIFF
--- a/src/sender.rs
+++ b/src/sender.rs
@@ -11,7 +11,7 @@ use rand::{rng, CryptoRng, Rng};
 use tracing::{debug, error, info, trace, warn};
 use tracing_futures::Instrument;
 use uuid::Uuid;
-use zkgroup::GROUP_IDENTIFIER_LEN;
+use zkgroup::{groups::GroupSendFullToken, GROUP_IDENTIFIER_LEN};
 
 use crate::{
     cipher::{get_preferred_protocol_address, ServiceCipher},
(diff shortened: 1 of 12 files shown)


rubdos:
Builds on #441, will clean this mess up a bit more after that merges.

rubdos:
Will now build on #454 and #464. I created `MultiRecipientAccess::GroupSendToken`, but this is currently exercised nowhere in favour of the (already legacy) UAK-xor-shenennigan.

src/sender.rs
```diff
@@ -582,174 +704,689 @@ where
...
+        unidentified: bool,
+        needs_sync: bool,
+    ) -> Result<SentMessage, MessageSenderError> {
+        let used_identity_key = self
+            .protocol_store
+            .get_identity(&recipient.to_protocol_address(*DEFAULT_DEVICE_ID)?)
+            .await?
+            .ok_or(MessageSenderError::UntrustedIdentity {
+                address: recipient,
+            })?;
```
rubdos, line 1050: This might trigger identity resets after successful sends. IIRC we have an identity not found error as well, but we might even want to `expect` it here, or just pass in the identity.

rubdos: Or just a plain `ProtocolError`; probably same issue in the `MismatchedDevices` recovery loop

rubdos:
> Tested working, but I still need a few passes through the code to get rid of the "vibes".

This branch is a complete (in large vibe-coded) mess, but a *working* mess. I'll start opening separate PRs for smaller reviewable blocks now. This is gonna take a bit longer than I hoped.

PR #454, opened 2026-07-04
Multi recipient endpoint

### LINKED ISSUE #451: Expose and use /v1/messages/multi_recipient in sender key sending

rubdos:
Fixes #451 

Builds on #464, will undraft when that's in.


rubdos:
Folded in #453

PR #456, opened 2026-07-05
Update link capabilities

Schmiddiii:
According to the fields of Signal-Desktop.


direc85 (approved):
This fixes HTTP 400 when registering as primary.

Edit: This is not correct; I think I tried two changes at the same go and drew false conclusions. (Will try linking next anyway so we'll find out soon.)

PR #430, opened 2026-04-26
Prekey refactor

rubdos:
Introduces methods to *replace* old prekeys. Implementors should:

1. Take care to remove all but the supplied one-time prekeys
2. Only replenish prekeys after receiving the QueueEmpty signal, or better: mark "deleted" prekeys as stale and remove after some hours.

Not in this PR: prekey staleness management.

This is mostly done by some cheap LLMs to play around, so I'll have to reread everything a few times to make sure it's kinda sound.

PR #457, opened 2026-07-05
Make registration work again

direc85:
Tested with Whisperfish to Signal Staging servers.

Continued from what @gferon did earlier.

### DIFF
--- a/src/websocket/account.rs
+++ b/src/websocket/account.rs
@@ -51,8 +51,6 @@ pub(crate) struct DeviceInfoEncrypted {
 #[serde(rename_all = "camelCase")]
 /// kept in sync with https://github.com/signalapp/Signal-Server/blob/main/service/src/main/java/org/whispersystems/textsecuregcm/entities/AccountAttributes.java#L25
 pub struct AccountAttributes {
-    // #[serde(default, skip_serializing_if = "Option::is_none")]
-    // pub signaling_key: Option<String>,
     pub registration_id: u32,
     pub voice: bool,
     pub video: bool,
--- a/src/websocket/account.rs
+++ b/src/websocket/account.rs
@@ -51,19 +51,23 @@ pub(crate) struct DeviceInfoEncrypted {
 #[serde(rename_all = "camelCase")]
 /// kept in sync with https://github.com/signalapp/Signal-Server/blob/main/service/src/main/java/org/whispersystems/textsecuregcm/entities/AccountAttributes.java#L25
 pub struct AccountAttributes {
-    pub fetches_messages: bool,
+    // #[serde(default, skip_serializing_if = "Option::is_none")]
+    // pub signaling_key: Option<String>,
     pub registration_id: u32,
-    pub pni_registration_id: u32,
-    #[serde(default, with = "serde_optional_prost_base64")]
-    pub name: Option<DeviceName>,
+    pub voice: bool,
+    pub video: bool,
+    pub fetches_messages: bool,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub registration_lock: Option<String>,
     #[serde(default, with = "serde_optional_base64")]
     pub unidentified_access_key: Option<Vec<u8>>,
     pub unrestricted_unidentified_access: bool,
-    pub capabilities: DeviceCapabilities,
     pub discoverable_by_phone_number: bool,
-    pub pin: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub capabilities: Option<DeviceCapabilities>,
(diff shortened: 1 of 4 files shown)


src/websocket/registration.rs
```diff
@@ -229,50 +250,51 @@ impl SignalWebSocket<websocket::Unidentified> {
         .json()
         .await
     }
-}
 
-impl SignalWebSocket<websocket::Identified> {
     pub async fn submit_registration_request(
         &mut self,
         registration_method: RegistrationMethod<'_>,
+        phonenumber: &str,
```
Schmiddiii: Nitpick: Can we maybe use the `PhoneNumber` type?

Same also in `register_account`.

src/websocket/account.rs
```diff
@@ -51,19 +51,23 @@ pub(crate) struct DeviceInfoEncrypted {
 #[serde(rename_all = "camelCase")]
 /// kept in sync with https://github.com/signalapp/Signal-Server/blob/main/service/src/main/java/org/whispersystems/textsecuregcm/entities/AccountAttributes.java#L25
 pub struct AccountAttributes {
-    pub fetches_messages: bool,
+    // #[serde(default, skip_serializing_if = "Option::is_none")]
+    // pub signaling_key: Option<String>,
```
Schmiddiii: Nitpick: Can probably be removed.

Schmiddiii (approved):
Looks good to me overall.

src/websocket/account.rs
rubdos: Agree, let's get rid of this (finally!!)

src/websocket/registration.rs
```diff
@@ -69,6 +72,17 @@ impl<'a> RegistrationMethod<'a> {
...
+    pub pni_identity_key: Vec<u8>,
+    pub aci_signed_pre_key: SignedPreKeyEntity,
+    pub pni_signed_pre_key: SignedPreKeyEntity,
+    pub aci_pq_last_resort_pre_key: KyberPreKeyEntity,
+    pub pni_pq_last_resort_pre_key: KyberPreKeyEntity,
+}
+
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct DeviceActivationRequest {
```
rubdos, line 88: This struct is dead now, right?

src/websocket/registration.rs
```diff
@@ -229,50 +250,51 @@ impl SignalWebSocket<websocket::Unidentified> {
...
-                pni_identity_key: pni_identity_key.serialize().into(),
-                device_activation_request,
-                every_signed_key_valid: true,
+                aci_identity_key: keys.aci_identity_key,
+                pni_identity_key: keys.pni_identity_key,
+                aci_signed_pre_key: keys.aci_signed_pre_key,
+                pni_signed_pre_key: keys.pni_signed_pre_key,
+                aci_pq_last_resort_pre_key: keys.aci_pq_last_resort_pre_key,
+                pni_pq_last_resort_pre_key: keys.pni_pq_last_resort_pre_key,
+                gcm_token: None,
```
rubdos: We should also expose this as `submit_registration_request` argument

src/websocket/registration.rs
```diff
@@ -304,4 +326,91 @@ impl SignalWebSocket<websocket::Identified> {
...
+            aci_last_resort_kyber_prekey,
+        ) = crate::pre_keys::replenish_pre_keys(
+            aci_protocol_store,
+            csprng,
+            &aci_identity_key_pair,
+            true,
+            0,
+            0,
+        )
+        .await?;
```
rubdos, line 375: Oof this is going to conflict with #430. But let's go ahead for now.

src/websocket/registration.rs
```diff
@@ -304,4 +326,91 @@ impl SignalWebSocket<websocket::Identified> {
...
+        Pni: PreKeysStore + IdentityKeyStore,
+    >(
+        &mut self,
+        csprng: &mut R,
+        registration_method: RegistrationMethod<'_>,
+        account_attributes: AccountAttributes,
+        aci_protocol_store: &mut Aci,
+        pni_protocol_store: &mut Pni,
+        skip_device_transfer: bool,
+        phonenumber: &str,
```
rubdos: PhoneNumber here as well (our internal `impl Into<utils::E164>` type, I'd say)

rubdos (changes requested):
Good stuff! I agree on @Schmiddiii's comments, and I have a few more. All easy fixes.

src/websocket/registration.rs
rubdos: It is not!

rubdos (approved):

PR #460, opened 2026-07-23
build(deps): upgrade libsignal to 0.99.0

boxdot:


Schmiddiii (approved):
Seems like you mixed a bit of #459 into this MR, but LGTM skimming over it.

boxdot:
> Seems like you mixed a bit of #459 into this MR, but LGTM skimming over it.

I struggled to base this pr on the former. Let's merge #459 first, and then I will rebase.

src/groups_v2/operations.rs
```diff
@@ -861,14 +861,14 @@ impl GroupOperations {
     // `PRESENTATION_VERSION_4`) without verifying that the Signal server accepts
     // it. A mismatched version will cause ZK proof verification to fail and
     // members will be rejected when joining groups.
```
rubdos: I think this comment needs an update; does libsignal now not expose any old version anymore?

rubdos (commented):
Omg, we finally got rid of the rand shim <3

LGTM, but that one comment probably needs a bump.

src/groups_v2/operations.rs
boxdot: Good point. The comment does not apply anymore. Removed.

gferon (approved):

PR #462, opened 2026-07-25
Rework pre-key management, lifecycle, tracking, rotation

direc85:
As quite obvious, I used Claude heavily making this PR. I think I get every bit and piece individually, but as I'm not an expert in Signal innards, I can't judge the bigger picture. This majorly reworks pre-key handling, and I'm happy to see strictness in several places now. It feels like we've been riding through the night with no headlights, without major accidents so far, and we *really* can't keep doing that. I have been running this in my daily device for some time now and nothing has blown up yet. It seems to work, but please do review with keen eye for slop. And stupid.

Without further ado...

# Rework pre-key management: stale-then-delete lifecycle, persisted ID tracking, split rotation

## Summary

Reworks pre-key generation, storage and rotation to fix ID reuse and align with Signal Android. The old flow persisted keys inside generation, derived next-IDs from `MAX(id)`, and replaced one-time keys wholesale — causing repeated-key 409s and deleting still-published keys mid-handshake.

Two separations do the work. **Generation, persistence and ID bookkeeping** are pulled apart: generation is pure, persistence is dumb, and all ID advancement and active-ID recording is explicit. **Signed and one-time rotation** are split onto independent triggers: the signed and last-resort keys rotate on an age schedule, while one-time keys are replenished only when the server reports it is running low.

## Motivation

1. **ID reuse → 409.** Next-IDs came from `MAX(id)`. Deleting the highest key let the next allocation reuse a freed ID; the server rejected it, and re-upload looped because the active signed-prekey ID was never advanced.
2. **Wholesale replace deleted live keys.** One-time keys were deleted and reinserted on every refresh, removing keys the server might still hand out.
3. **No active-ID tracking.** The store called the highest stored ID "active" rather than the one the server serves.
4. **Unconditional churn.** Every refresh uploaded a full batch of one-time keys regardless of consumption.

### Trait surface (`PreKeysStore`)

New `SignedPreKeyStoreExt` supertrait (`load_signed_pre_keys`, `remove_signed_pre_key`). `PreKeysStore` gains next-ID setters, active-ID setters (getter `signed_prekey_id` → `active_signed_prekey_id`), `ec_one_time_pre_keys_count`, the EC staleness hooks, and `last_prekey_rotation` / `set_last_prekey_rotation` for the rotation schedule.

The composed operations are **default methods** on the trait rather than free functions: `generate_signed_pre_keys`, `generate_one_time_pre_keys`, `store_signed_pre_key_bundle`, `store_one_time_pre_key_bundle`, `mark_signed_pre_keys_active`, `clean_stale_pre_keys`, `signed_pre_keys_due_for_rotation`. Implementors get them free; only the persistence primitives are required.

Two contracts are documented on the trait: setters only persist (the advance and wrap are computed by the caller), and `set_active_*` runs only after a successful upload.

### Rotation split (`account_manager.rs`)

`update_pre_key_bundle` becomes two independently gated operations:

- `rotate_signed_pre_keys` — age-gated (2 days), uploads one signed key and one last-resort Kyber key, leaving the one-time pool alone.
- `replenish_one_time_pre_keys` — queries the server's remaining count and uploads a batch only below `PRE_KEY_MINIMUM`, so a routine refresh usually uploads nothing here.

`PreKeyState`'s sections are now optional, so each upload carries only what it replaces. Confirmed on a staging server that a PUT omitting `preKeys`/`pqPreKeys` leaves the server's one-time pool intact.

### Ordering — deliberate divergence from Android

Next-IDs advance pre-upload (a gap on failure is harmless). Active IDs are recorded **post-upload only**, so cleanup never preserves a key the server hasn't accepted, nor deletes the key it still serves. Android sets active pre-upload; this errs safer.

### Removals

`pnp_initialize_devices` is deleted — a one-shot pre-PNI migration no live account can reach, since any account old enough to need it has long since expired server-side.

## Breaking changes

- `PreKeysStore` implementors must add the new required methods; the composed operations arrive as defaults. `ExampleStore` is updated with `todo!()` stubs.
- `signed_prekey_id` → `active_signed_prekey_id`.
- `register_pre_keys` takes `&PreKeyState`; its one-time and signed sections are now `Option`.
- `replenish_pre_keys` removed — use the generate/store pairs, or `update_pre_key_bundle`.
- `pnp_initialize_devices` removed.

## Caller responsibility ⚠️

Registration and linking upload signed and last-resort keys only, so a new device has no one-time keys on the server until its first `update_pre_key_bundle`. Callers should also drain the message queue before rotating: persistence isn't atomic with message handling, though the 90-day stale grace period and `min_count` floor mitigate it.

## Testing

Store-level tests live in the Whisperfish implementation: next-ID no-regress after delete (the 409 bug), the stale→delete lifecycle with a matrix pinning the retention arithmetic, signed rotation leaving one-time pools untouched, and per-identity ID-space independence. Constants match Signal Android master (`BATCH_SIZE` 100, `ARCHIVE_AGE` 30d, `ONE_TIME_PREKEY_MINIMUM` 10, `REFRESH_INTERVAL` 2d), as does the stale-marking guard.

## Follow-ups

Android gates EC and Kyber replenishment independently; this gates them together. No `forceRotation` equivalent, and no `MAXIMUM_ALLOWED_SIGNED_PREKEY_AGE` failsafe — upstream blocks sending and rotates synchronously when a signed key is 14 days stale or the elapsed time is negative.

### DIFF
--- a/src/pre_keys.rs
+++ b/src/pre_keys.rs
@@ -14,7 +14,6 @@ use libsignal_protocol::{
 
 use rand::{CryptoRng, Rng};
 use serde::{Deserialize, Serialize};
-use tracing::Instrument;
 
 #[async_trait(?Send)]
 /// Additional methods for the Kyber pre key store
@@ -50,17 +49,48 @@ pub trait KyberPreKeyStoreExt: KyberPreKeyStore {
     ) -> Result<(), SignalProtocolError>;
 }
 
+#[async_trait(?Send)]
+/// Additional methods for the signed pre key store
+pub trait SignedPreKeyStoreExt: SignedPreKeyStore {
+    async fn load_signed_pre_keys(
+        &self,
+    ) -> Result<Vec<SignedPreKeyRecord>, SignalProtocolError>;
+
+    async fn remove_signed_pre_key(
+        &self,
+        pre_key_id: SignedPreKeyId,
+    ) -> Result<(), SignalProtocolError>;
+}
+
 /// Stores the ID of keys published ahead of time
 ///
 /// <https://signal.org/docs/specifications/x3dh/>
+///
+/// ## Next-ID advance contract
+///
+/// Implementors of the `set_next_*` setters MUST treat them as plain
+/// persistence: write the value, return. They MUST NOT advance, wrap, or
+/// otherwise mutate the id. All advancing (increment + `% PRE_KEY_MEDIUM_MAX_VALUE`
+/// wrap, starting from 1) is performed by libsignal-service-rs orchestration after key
+/// generation, then written via the setter. The `store_one_time_*` methods
+/// likewise only persist records; they do NOT advance next-ids.
+///
+/// ## Active-ID contract
+///
+/// `set_active_*` records the id the server has accepted. Orchestration calls
+/// these only after a successful upload, so `clean_*` excludes the correct
+/// live key. The matching getters return `None` before the first upload.
 #[async_trait(?Send)]
 pub trait PreKeysStore:
     PreKeyStore
     + IdentityKeyStore
     + SignedPreKeyStore
+    + SignedPreKeyStoreExt
     + KyberPreKeyStore
     + KyberPreKeyStoreExt
 {
+    // ---- Next-ID getters ----
+
     /// ID of the next pre key
     async fn next_pre_key_id(&self) -> Result<u32, SignalProtocolError>;
 
@@ -69,6 +99,25 @@ pub trait PreKeysStore:
 
     /// ID of the next PQ pre key
     async fn next_pq_pre_key_id(&self) -> Result<u32, SignalProtocolError>;
+
+    // ---- Next-ID setters (persistence only; do NOT advance) ----
+
+    async fn set_next_pre_key_id(
+        &mut self,
+        id: u32,
+    ) -> Result<(), SignalProtocolError>;
+
+    async fn set_next_signed_pre_key_id(
+        &mut self,
+        id: u32,
+    ) -> Result<(), SignalProtocolError>;
+
+    async fn set_next_pq_pre_key_id(
+        &mut self,
+        id: u32,
+    ) -> Result<(), SignalProtocolError>;
+
+    // ---- Counts ----
 
     /// number of signed pre-keys we currently have in store
     async fn signed_pre_keys_count(&self)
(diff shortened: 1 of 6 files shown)


direc85:
Includes #469

Edit: Not anymore :)

src/pre_keys.rs
```diff
@@ -306,3 +374,275 @@ pub(crate) async fn replenish_pre_keys<R: Rng + CryptoRng, P: PreKeysStore>(
...
+    Ok(())
+}
+
+const ARCHIVE_AGE: chrono::Duration = chrono::Duration::days(30);
+const STALE_AGE: chrono::Duration = chrono::Duration::days(90);
+const ONE_TIME_MIN_COUNT: usize = 200;
+
+pub async fn clean_signed_pre_keys<P: PreKeysStore>(
+    protocol_store: &mut P,
+) -> Result<(), SignalProtocolError> {
```
rubdos: These, and the other loose functions, are rather LLM-isms. I would make these default methods on the PreKeysStore trait (which is why it exists in the first place). For the loose methods where we'd have to change the libsignal_protocol traits: extension traits.

src/pre_keys.rs
```diff
@@ -306,3 +374,275 @@ pub(crate) async fn replenish_pre_keys<R: Rng + CryptoRng, P: PreKeysStore>(
...
+    let Some(active_id) = protocol_store.active_signed_prekey_id().await?
+    else {
+        tracing::debug!("No active signed pre-key; nothing to clean");
+        return Ok(());
+    };
+
+    let now_ms = chrono::Utc::now().timestamp_millis() as u64;
```
rubdos: At first sight, these millis calls (also in other impls) are used for comparison purposes. No need to convert to millis if we can just use the actual chrono (timezoned, Utc) comparison methods.

src/pre_keys.rs
```diff
@@ -306,3 +374,275 @@ pub(crate) async fn replenish_pre_keys<R: Rng + CryptoRng, P: PreKeysStore>(
...
+/// cleanup run. On upload failure, next-ids stay advanced (harmless gap) and the
+/// previously-active keys remain live.
+pub async fn rotate_pre_keys<P, R, F, Fut>(
+    protocol_store: &mut P,
+    csprng: &mut R,
+    identity_key_pair: &IdentityKeyPair,
+    use_last_resort_key: bool,
+    pre_key_count: u32,
+    kyber_pre_key_count: u32,
+    upload: F,
```
rubdos: This is `Fut: ...` + `F: Fn() -> Fut` thing is also an LLM-ism imo (I've seen GLM do this a lot...)

In practice, the upload fun+future is always an instance of a concrete type here, so I'd honestly rather see the concrete type passed in.

src/pre_keys.rs
```diff
@@ -306,3 +374,275 @@ pub(crate) async fn replenish_pre_keys<R: Rng + CryptoRng, P: PreKeysStore>(
...
+        tracing::info!(?id, ts, "removing old signed pre-key");
+        protocol_store.remove_signed_pre_key(id).await?;
+    }
+
+    Ok(())
+}
+
+pub async fn clean_last_resort_kyber_pre_keys<P: PreKeysStore>(
+    protocol_store: &mut P,
+) -> Result<(), SignalProtocolError> {
```
rubdos: Another thing I'm noticing is that all these cleaning methods are the exact same thing (fetch 'em all, compare the timestamp, prune what's too old with maybe a threshold). Can we somehow abstract the logic out?

rubdos (changes requested):
This is purely my Rust brain looking at the code now. I'll need my Signal brain in another pass.

src/pre_keys.rs
direc85: That's a dead function near identical to `update_pre_key_bundle` and was already dropped mid-review.

src/pre_keys.rs
direc85: Sounds like a good idea, noticed the same but didn't want to fight Diesel at the time...

src/pre_keys.rs
direc85: That is slightly cleaner that way indeed.

src/pre_keys.rs
direc85: I tried this, but it would just move the point of conversion elsewhere, or would make us use `Timestamp` throughout, which quickly spirals out of control. IMO not worth it.

src/pre_keys.rs
direc85: Got something done; not quite as clean as I'd hoped, but better.

PR #464, opened 2026-08-01
feat: error handling specialization

rubdos:
Picks my error handling idea from #454

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -75,6 +75,21 @@ pub fn parse_pni_with_fallback(
     })
 }
 
+/// Deserialize a [`ServiceId`] from its `"ACI:uuid"` / `"PNI:uuid"` server
+/// string form.
+pub fn deserialize_service_id<'de, D>(
+    deserializer: D,
+) -> Result<ServiceId, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    use serde::Deserialize;
+    let s = <&'de str>::deserialize(deserializer)?;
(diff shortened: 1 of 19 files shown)


rubdos:
One follow-up that I would consider, is to make ServiceError generic, depending on which request-specific errors we trigger. Or invert it: return more specific error types from the request methods, which have ServiceError as a variant.

src/utils.rs
```diff
@@ -75,6 +75,21 @@ pub fn parse_pni_with_fallback(
     })
 }
 
+/// Deserialize a [`ServiceId`] from its `"ACI:uuid"` / `"PNI:uuid"` server
+/// string form.
```
rubdos, line 79:
```suggestion
/// Deserialize a [`ServiceId`] from its `"uuid"` / `"PNI:uuid"` string form.
```

Schmiddiii (commented):
Skimmed a little. I certainly don't understand the `error_mapper` macro code, but what it does seems intuitively clear. Adding all those additional error codes is also a great idea. Overall looks good.

direc85 (approved):
Yeah that macro is quite something, but everything else seems nice and clean now. LGTM from me, assuming you have given this a spin already.

rubdos:
> Yeah that macro is quite something, but everything else seems nice and clean now. LGTM from me, assuming you have given this a spin already.

Been dailying this including the very vibey #453 for two days. That latter one needs serious work though; that's in a bad condition.

PR #465, opened 2026-08-03
feat: validate and expose PNI signatures

rubdos:
Some slight refactors and cleanups to make this happen.

### DIFF
--- a/src/cipher.rs
+++ b/src/cipher.rs
@@ -90,6 +90,18 @@ where
     /// Opens ("decrypts") an envelope.
     ///
     /// Envelopes may be empty, in which case this method returns `Ok(None)`
+    ///
+    /// # PNI signature side-cars
+    ///
+    /// A side-car is verified during decryption and, when valid, its
+    /// PNI is exposed as `Metadata::pni_verified`. A side-car on a message with
+    /// no `content` is discarded, with a `tracing::warn!`; upstream Signal
+    /// clients only ever attach a side-car alongside `content`, so this path is
+    /// unreachable in practice, and wiring it through would change this method's
+    /// return type.
+    ///
+    /// **NOTE**: must process `Metadata::pni_verified` to confirm the sender's
+    /// PNI, not the raw side-car message.
     #[tracing::instrument(skip(envelope, csprng), fields(envelope = debug_envelope(&envelope)))]
     pub async fn open_envelope<R: Rng + CryptoRng>(
         &mut self,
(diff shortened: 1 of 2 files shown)


src/cipher.rs
```diff
@@ -107,6 +108,15 @@ where
                 .record("envelope_metadata", plaintext.metadata.to_string());
 
             let Some(content) = &plaintext.message.content else {
+                // Cheap-out: post-decrypt side-ops are not propagated alongside the
+                // Option<Content>, so a content-less message drops its sidecar here.
+                // TODO: return a Vec<DecryptPostOp> next to the Option<Content>.
```
direc85, line 125: Is this something we need before merge?

Shower thought: could we pass some null message instead of None? That would be a hack but maybe enough until the second use case appears?

direc85 (commented):
Pending approve, depending on the sole comment I left. This is still a draft, so... :)

src/cipher.rs
rubdos: No, because it doesn't add immediate value (given the constraints documented), complicates the API (quite a bit, for the consumer of the API), and we don't have anything other current variants implemented for the `DecryptPostOp` enum... My vote is 100% postpone :')

direc85: Sure, agreed.

Another thought: Do we know what kind of messages get sent when a contact re-registers? If it's a message with PNI signature only, without body, wouldn't that get dropped then? (I think I was after that already.)

rubdos: Three things:

1. pni-sig only messages do not currently exist; and
2. re-registration doesn't trigger any special messages; but
3. your comment made me think what *does* happen on re-reg.

On re-reg, the contact record might need to be split: if an ACI is assumed unregistered, its PNI might not and needs to be split from the ACI into a new recipient:

> StorageService ContactRecord carries the ACI/PNI/e164 triple. ContactRecordProcessor even splits the pair during storage sync — for unregistered ACI-only
> records (no PNI, no e164) it calls splitForStorageSyncIfNecessary(aci), which nulls PNI/E164 on that row and re-merges the PNI into its own recipient
> row. The code comment is explicit:
>│ "we want to split unregistered users into separate rows so that a user could re-register and get a different ACI."
>│ (ContactRecordProcessor.kt:64, RecipientTable.kt:2612)

This all boils down to phone number changes too, but I'm afraid that's out of scope here (both in this PR and even in the lib), and it's quite complicated.

src/content.rs
```diff
@@ -26,6 +25,9 @@ pub struct Metadata {
     pub sender: ServiceId,
     pub destination: ServiceId,
     pub sender_device: DeviceId,
+    /// Contains the PNI of the sender if verification was succesful
```
rubdos:
```suggestion
    /// Contains the PNI of the sender if verification was succesfull
```

PR #467, opened 2026-08-16
Expose attachment download metadata

d3vv3:
## Summary

- return CDN response metadata with attachment streams
- expose the optional Content-Length for progress reporting
- preserve the existing stream-only API for current callers

## Verification

- `cargo fmt --check`
- `git diff --check upstream/main...HEAD`
- local build/test reached `boring-sys` but the host Clang environment lacks `stddef.h`; no project test failed

### DIFF
--- a/src/push_service/cdn.rs
+++ b/src/push_service/cdn.rs
@@ -5,7 +5,7 @@ use std::{
 
 use futures::TryStreamExt;
 use reqwest::{
-    header::{CONTENT_LENGTH, CONTENT_RANGE, CONTENT_TYPE, RANGE},
+    header::{HeaderMap, CONTENT_LENGTH, CONTENT_RANGE, CONTENT_TYPE, RANGE},
     multipart::Part,
     Method, StatusCode,
 };
@@ -55,11 +55,30 @@ pub struct ResumeInfo {
     pub content_start: u64,
 }
 
+pub struct AttachmentDownload<R> {
+    pub stream: R,
+    pub content_length: Option<u64>,
+}
+
+fn content_length(headers: &HeaderMap) -> Option<u64> {
+    headers.get(CONTENT_LENGTH)?.to_str().ok()?.parse().ok()
+}
+
 impl PushService {
     pub async fn get_attachment(
         &mut self,
         ptr: &AttachmentPointer,
     ) -> Result<impl futures::io::AsyncRead + Send + Unpin, ServiceError> {
+        Ok(self.get_attachment_with_metadata(ptr).await?.stream)
+    }
+
+    pub async fn get_attachment_with_metadata(
+        &mut self,
+        ptr: &AttachmentPointer,
+    ) -> Result<
+        AttachmentDownload<impl futures::io::AsyncRead + Send + Unpin>,
+        ServiceError,
+    > {
         let path = match ptr.attachment_identifier.as_ref() {
             Some(AttachmentIdentifier::CdnId(id)) => {
                 format!("attachments/{}", id)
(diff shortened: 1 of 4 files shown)


src/push_service/cdn.rs
```diff
@@ -55,11 +55,30 @@ pub struct ResumeInfo {
...
 }
 
+pub struct AttachmentDownload<R> {
+    pub stream: R,
+    pub content_length: Option<u64>,
+}
+
+fn content_length(headers: &HeaderMap) -> Option<u64> {
+    headers.get(CONTENT_LENGTH)?.to_str().ok()?.parse().ok()
+}
```
rubdos: Inline this function, and at least `trace::warn!` a warning on parse-failure.

src/push_service/cdn.rs
```diff
@@ -55,11 +55,30 @@ pub struct ResumeInfo {
...
+        Ok(self.get_attachment_with_metadata(ptr).await?.stream)
+    }
+
+    pub async fn get_attachment_with_metadata(
+        &mut self,
+        ptr: &AttachmentPointer,
+    ) -> Result<
+        AttachmentDownload<impl futures::io::AsyncRead + Send + Unpin>,
+        ServiceError,
+    > {
```
rubdos: Just break the API, we're not gonna bother maintaining two variants here and it's a trivial break/fix. If you *really **really*** want, you could `impl AsyncRead for AttachmentDownload<R: AsyncRead>`.

src/push_service/cdn.rs
```diff
@@ -82,20 +102,37 @@ impl PushService {
         cdn_id: u32,
         path: &str,
     ) -> Result<impl futures::io::AsyncRead + Send + Unpin, ServiceError> {
-        let response_stream = self
+        Ok(self.get_from_cdn_with_metadata(cdn_id, path).await?.stream)
+    }
+
+    #[tracing::instrument(skip(self))]
+    pub(crate) async fn get_from_cdn_with_metadata(
```
rubdos: Again, just break API.

src/push_service/cdn.rs
```diff
@@ -440,3 +477,52 @@ impl PushService {
...
+
+#[cfg(test)]
+mod tests {
+    use futures::io::Cursor;
+    use reqwest::header::{HeaderMap, HeaderValue, CONTENT_LENGTH};
+
+    use super::{content_length, AttachmentDownload};
+
+    #[test]
+    fn parses_content_length_header() {
```
rubdos: This is a waste of tokens. You're testing HeaderMap's ability for parsing, not your code.

src/push_service/cdn.rs
```diff
@@ -440,3 +477,52 @@ impl PushService {
...
+
+        headers.insert(
+            CONTENT_LENGTH,
+            HeaderValue::from_bytes(b"\xff").expect("valid header value"),
+        );
+        assert_eq!(content_length(&headers), None);
+    }
+
+    #[test]
+    fn attachment_download_stream_is_compatible_with_legacy_return_type() {
```
rubdos: Even more waste of tokens. Just remove both tests, they don't prove anything.

rubdos (changes requested):
The feature is valuable, but this is very AI-sloppy as-is.

d3vv3:
Thanks for your comments. I am no expert in Rust so LLMs are an enabler for me in such cases. I hope it is better now, looks like so.

rubdos:
> I am no expert in Rust so LLMs are an enabler for me in such cases. I hope it is better now, looks like so.

I find that the most dangerous use case for agentic coding, but it indeed looks in good shape now. @direc85 said LGTM out of band on this version too; merging!

PR #468, opened 2026-08-29
fix: process SKDMs before dropping content-less message

rubdos:


direc85 (approved):
Passed the smoke test and seems to be a much needed imprvement -> LGTM

PR #471, opened 2026-09-04
refactor(sender): extract repair_recipient_sessions from 409/410 handling

rubdos:
Makes reusing and repurposing this easier in #453; was easy to pick out.

### DIFF
--- a/src/sender.rs
+++ b/src/sender.rs
@@ -608,8 +608,6 @@ where
 
         let content_bytes = content.encode_to_vec();
 
-        let mut rng = rng();
-
         for _ in 0..4u8 {
             let Some(EncryptedMessages {
                 messages,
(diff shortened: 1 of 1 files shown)


src/sender.rs
```diff
@@ -750,6 +703,61 @@ where
...
+    /// and establish sessions with devices we did not know about
+    /// (`missing_devices`), so the caller can rebuild and retry the send.
+    async fn repair_recipient_sessions(
+        &mut self,
+        recipient: ServiceId,
+        extra_devices: &[DeviceId],
+        missing_devices: &[DeviceId],
+    ) -> Result<(), MessageSenderError> {
+        for device_id in extra_devices {
+            tracing::debug!("dropping session with device {}", device_id);
```
rubdos, line 718: This should be a field, not a log line. Probably also trace-level.

src/sender.rs
```diff
@@ -750,6 +703,61 @@ where
...
+            self.protocol_store
+                .delete_service_addr_device_session(
+                    &recipient.to_protocol_address(*device_id)?,
+                )
+                .await?;
```
rubdos, line 723: Should we do best-effort: exhaust the inputs, and return a Vec of errors?

src/sender.rs
```diff
@@ -750,6 +703,61 @@ where
...
+        }
+
+        let mut rng = rng();
+
+        for device_id in missing_devices {
+            tracing::debug!(
+                "creating session with missing device {}",
+                device_id
+            );
```
rubdos, line 732: Same as above: field, or even a *span* during the whole loop (although don't hold across await...) makes way more sense

src/sender.rs
```diff
@@ -750,6 +703,61 @@ where
...
+                    .expect("valid device id"),
+                &mut self.protocol_store.clone(),
+                &mut self.protocol_store,
+                &pre_key,
+                SystemTime::now(),
+                &mut rng,
+            )
+            .await
+            .map_err(|e| {
+                error!(%e, ?recipient, "failed to create session");
```
rubdos, line 753:
```suggestion
                tracing::error!(%e, ?recipient, "failed to create session");
```

I prefer fully-qualified here; might be confusing with `anyhow::error!`.

PR #472, opened 2026-09-08
Multi recipient API

### LINKED ISSUE #451: Expose and use /v1/messages/multi_recipient in sender key sending

rubdos:
Picks the multi-recipient API code from #453. Together with #471 these should be the easiest parts to review and merge... (This also technically replaces #454)

Fixes #451

### DIFF
--- a/src/sender.rs
+++ b/src/sender.rs
@@ -11,7 +11,7 @@ use rand::{rng, CryptoRng, Rng};
 use tracing::{debug, error, info, trace, warn};
 use tracing_futures::Instrument;
 use uuid::Uuid;
-use zkgroup::GROUP_IDENTIFIER_LEN;
+use zkgroup::{groups::GroupSendFullToken, GROUP_IDENTIFIER_LEN};
 
 use crate::{
     cipher::{get_preferred_protocol_address, ServiceCipher},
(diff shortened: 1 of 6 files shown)


src/sender.rs
```diff
@@ -70,6 +72,40 @@ pub struct SentMessage {
...
+/// For non-story multi-recipient sends Signal requires **exactly one** of:
+/// - a combined unidentified-access key (the bitwise XOR of every
+///   recipient's 16-byte UAK), sent base64-encoded in the
+///   `Unidentified-Access-Key` header, or
+/// - a group-send endorsement full token, sent base64-encoded in the
+///   `Group-Send-Token` header.
+///
+/// Stories carry neither header; pass `None` for `access`.
+#[derive(Clone)]
+pub enum MultiRecipientAccess {
```
rubdos, line 86: Let's move this to `unidentified_access.rs`

src/websocket/request.rs
```diff
@@ -52,6 +52,17 @@ impl WebSocketRequestMessageBuilder {
...
+    /// Attach a raw body (e.g. the Sealed Sender v2 multi-recipient MRM
+    /// payload) with an explicit `Content-Type` header.
+    pub fn body(
+        mut self,
+        content_type: &str,
+        body: Vec<u8>,
+    ) -> WebSocketRequestMessage {
+        self.request.body = Some(body);
+        self.header("content-type", content_type).request
+    }
```
rubdos, line 64: I doubt this method is necessary. Let's just call `header()` and use the existing `http_request` API, and maybe make a custom `send()` method for custom body strings (which we don't necessarily want to allocate, I'd say?

